PR #9, opened 2015-02-04
Use latest buildout-bootstrap.py. Add basic test.

tisto:
...ools-version=8.3 for now to work around issues with the latest setuptools version; Add test that creates a plone_addon and runs the tests inside it.


prakharjoshi:
hello,
         So is there a new bootstrap.py for all plone products or this is
for this one ? We also have to change the version of setup-tools for that
bootstrap ?

Thank You!!

Prakhar Joshi
DA-IICT,Gandhinagar

tisto:
This is the recommended way that is described on the pypi page of zc.buildout:

https://pypi.python.org/pypi/zc.buildout/2.3.1

We could rename bootstrap-buildout.py to bootstrap.py to keep the old way. Though, personally I perfer to keep in sync with what zc.buildout docs recommend.

The latest setuptools versions have issues, so we can't use them. In order to pin setuptools we have to upgrade to at least setuptools 8.3. Therefore we need a newer bootstrap-buildout.py. I know this is crazy btw. (but not our fault).

do3cc:
Can you add a comment somewhere why we need to pin to setuptools 8.3?

PR #13, opened 2015-02-04
Remove z2.installProducts. Use testing best practices.

tisto:
I think we can safely expect that we are not dealing with old zope2 only packages in 2015.

### DIFF
--- a/buildout.cfg
+++ b/buildout.cfg
@@ -28,7 +28,7 @@ input = inline:
    python bootstrap-buildout.py --setuptools-version=8.3
    ${buildout:directory}/test.plone_addon/bin/buildout
    ${buildout:directory}/test.plone_addon/bin/test
-   rm -rf ${buildout:directory}/test.plone_addon/
+   # rm -rf ${buildout:directory}/test.plone_addon/
 output = ${buildout:directory}/bin/test
 mode = 755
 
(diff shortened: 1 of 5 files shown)


buildout.cfg
```diff
@@ -28,7 +28,7 @@ input = inline:
    python bootstrap-buildout.py --setuptools-version=8.3
    ${buildout:directory}/test.plone_addon/bin/buildout
    ${buildout:directory}/test.plone_addon/bin/test
-   rm -rf ${buildout:directory}/test.plone_addon/
+   # rm -rf ${buildout:directory}/test.plone_addon/
```
do3cc: Was this an accident?

tisto: Yeah. Kind of. For debugging I kept the generated code. Should not be committed though.

cedricmessiant:
Sadly, we still have to deal with such packages in 2015. Two examples : Products.membrane and collective.taskqueue.pasplugin...

djowett:
Um, will this affect uninstall profiles? I was trying to add a PR for uninstall profile & tests, but my uninstall method isn't being called. Is Zope2 required for that? (I'm looking at https://plone.org/products/collective-flowplayer/issues/5)

djowett:
Ok, my question still stands, but it's  ab2eb80 rather than this change that is killing uninstall profiles

tisto:
Sure. There might be cases where this is needed. Though, this does not justify to have that in all packages. We should just try to cover the most important use cases and try to keep the code as clean as possible. Otherwise we will end up with hundreds of packages that carry over this code lines without any good reason.

PR #14, opened 2015-02-04
Remove testing profile.

tisto:
I consider a testing profile an anti-pattern for almost all use cases. It introduces a testing fixture for the entire package even though for most of the cases it is just used for a single test. This makes the testing setup brittle and should be avoided if possible.


pbauer:
:+1: kick it out!

PR #19, opened 2015-02-05
Just link to docs.plone.org. All other links might break in the future.

tisto:


do3cc:
I do not agree here. cant we add proper cross references in docs.plone.org to which we then link @svx

PR #25, opened 2015-02-06
Add includeDependencies.

tisto:


gforcada:
These "prepare release" and "back to development" should go directly to master right?

Can we squash together the other commits were it make sense also? :)

gforcada:
Apart from that it looks good of course!

Issue #10, opened 2015-02-04
Buildout configuration is spread over different files.


tisto:
While it is generally a good idea to have a buildout.d directory with multiple files for a deployment buildout, I don't think this makes any sense for a Plone add-on product. The buildout in an add-on product will always be quite simple and this has the potential to confuse people. 

I plan to use bobtemplates.plone for a training and I would prefer if I wouldn't to have to explain the entire buildout.d structure.

I would propose to merge all buildout.d files into a single top-level buildout.cfg. We can always create a second template for a deployment buildout for instance, where a buildout.d directory would make sense. Opinions?

BTW: buildout.d/jenkins.cfg is really really outdated and should not be used at all.

do3cc:
+1
Can you point to an uptodate jenkins.cfg?

tisto:
I'd be happy to create a newer jenkins.cfg. The question is do we want to recommend to use jenkins for Plone add-on products?

PR #30, opened 2015-02-12
Buildout simplified

tisto:


tisto:
See https://github.com/plone/bobtemplates.plone/issues/10

Issue #35, opened 2015-02-16
Create bobtemplate for Dexterity.


tisto:
So we can replace zopeskel here:

https://github.com/plone/documentation/blob/master/develop/addons/dexterity.rst

Related documentation issue: https://github.com/plone/documentation/issues/193#issuecomment-74475961

Question is: do we want to have dexterity/theme/etc as a separate bobtemplate or do we want to integrate it into the existing plone_addon template.

cc @svx

djowett:
Hi, my thoughts are
  mr.bob templates can overwrite each other's output (in a good way) pretty easily - the only reason they need to know about each other is to not clobber each other's output. (I believe paster/zopeskel localcommands were much more entangled in their "parents").

My suggestions:
- Since you may want to add dexterity behaviors / content types multiple times - then you are more or less *forced* to have a different bobtemplate
- In order to minimize clobbering worries - try to keep each new template working in different subfolders - e.g. this dexterity one in "content"

tisto:
@djowett Do you have an example of how to override the template output?

I was thinking about using different folders as well. Though, there will always be the need to register even a folder in configure.zcml and/or a gs setup step.

Things would be a lot easier if mr.bob would support depending on existing templates:

https://github.com/iElectric/mr.bob/issues/31

djowett:
It's very easy (unless I misunderstood):

    $ ../bin/mrbob -O collective.myaddon2 bobtemplates:plone_addon
    # I would recommend a git init ; git add . ; git commit here in real situations
    $ ../bin/mrbob -O collective.myaddon2 mrbob:template_sample/
    Welcome to mr.bob interactive mode. Before we generate directory structure, some questions need to be answered.

    Answer with a question mark to display help.
     Values in square brackets at the end of the questions show the default value if there is no answer.

     --> What is your name?: Daniel

     --> How old are you? [24]: 135

     --> Enter password: 

    Generated file structure at /home/daniel/dev/Plone/Development/highfields_ph2.1/src/collective.myaddon2
    $ ls -ltr collective.myaddon2 
     ....
     -rw-rw-r-- 1 daniel daniel   89 Feb 12 11:47 CHANGES.rst
     -rw-rw-r-- 1 daniel daniel   83 Feb 16 11:32 sample.txt

     $ more collective.myaddon2/sample.txt 
     My name is Daniel. I am 135 years old.
 
     This is my password to everything:  doodah

djowett:
[...]

Issue #29, opened 2015-02-12
Design Goals (Discussion)


tisto:
**Design Goals**
- Simple is better than complex.
- Cover 80% of use cases, don’t become too complex.
- Provide a good starting point for development, not a swiss army knife for the entire development cycle.
- Code and tests should reflect the current coding best practice.
- Generated code should have 100% test coverage.

**Concrete Decisions**
- Do not use local commands. They significantly increase complexity and mr.bob has no support for them anyways. We tried this with ZopeSkel/Templer and it did not work out that well.
- Limit the overall number of questions to not confuse people too much. Having separate packages or providing commented out code sections might be a good alternative.
- Fix problems in Plone itself rather than providing workarounds in our templates. (see https://github.com/plone/bobtemplates.plone/issues/27)
- Provide skeletons for the most important Plone package use cases, not for every possible scenario:
  - Add-on product (collective, plone, custom project, policy product)
  - Content Types (Dexterity, Archetypes?)
  - Theme

This is just a draft and open for discussion. Feel free to comment or propose additions. I would especially like to hear the opinion of @cewing and @smcmahon. If we can reach an agreement we can add those design goals to the docs.

aclark4life:
FWIW, IMHO:
- Kill the term "product" once and for all. Use terminology "package" to refer to "thing you add to the buildout" and "add-on" for "thing the package provides".
- PLIP bobtemplates.plone for inclusion in the Plone core so it gets "first class support". (Including it in the core does not guarantee it will be supported, but the best experience I've had with template/code generators is in Django/Pyramid where its obvious the core team put some thought in to the experience.)
- 100% code coverage for generated code is nice, but not strictly necessary. For the most part, anyone running a code generator is not going to care about test coverage for their generated code. And if you are generating enough code to require tests, something else is probably wrong and should be fixed in Plone itself (as per concrete decision 3).

do3cc:
Would it make sense to use branches to provide optional addons? That might make it simple to keep in sync with the base package.

pbauer:
* Focus on Plone 5 (no old-school js-registration, only Plone5-style). 
* Test against Plone 5
* +1 for shipping it with the installers
* We could even include it in Plone itself. There are two places where exporting customosations as a package is usefull:  
  * Dexterity: Additionally to "Export type-profile" we could "Create addon package with type"
  * Something better than https://github.com/plone/plone.app.themeeditor that exports customized templates/images/css/js as jbot-overides in a package
[...]

PR #46, opened 2015-02-21
Dexterity Support

tisto:
This pull requests adds dexterity support. Please do not merge since it is a work in progress!

Though, in order to proceed with this, we have to make a design decision regarding how we are going to ask questions to support the different use cases.

In order to keep the templates clean and testable, I removed quite a few unnecessary questions. For testablity it is essential that we keep things simple, because we just can not test every possible combination of questions there are. Therefore my idea was to initially ask the user, what kind of package she would like to create:

1) Basic
2) Dexterity
3) Theme

This allows us to test those three basic use cases with functional acceptance tests (run mr.bob with that specific configuration, run generated buildout, run tests for generated code to make sure everything works).

We already test like this on travis with a build matrix:

https://travis-ci.org/plone/bobtemplates.plone/

Depending on the answer of this first question, we can ask more question if needed for this specific use case and at the same time keep the number of questions small.

Though, we had a discussion on IRC yesterday and it turned out that some people prefer to put everything into one single package (Dexterity, Theme, etc.). This raises a few questions that I did not think about at first:

1) Do we consider putting everything in a single package as a valid use case?

2) Do we consider putting everything into a single package to be best practice?

3) Do we want to support that use case? E.g. if we decide that it is a valid use case but not considered best practice?

4) How are we going to support that use case?

I would like to hear opinions here.

Here is my personal opinion:

My main concern is, guess what, testability. For one thing I would consider putting everything into a single package to be an anti-pattern for testing. I won't go too much into detail here, but things are in general hard to test if they are not properly isolated. You will end up with a complex test fixture and your tests will take quite a while to run, making it impossible to do TDD. It might be ok for smaller projects, but I think we would send the wrong message and confusing people, if our best practice would be "it depends". Especially in smaller projects it is quite easy to keep the separation between theme, policy, content types. In larger projects, package architecture is hard. You need to put effort into decoupling packages and code. Though, I don't think our templates target large and complex projects.

Another important issue to put into considerations is the testability of our templates. Our acceptance tests are expensive to run, so we just can not test every single combination of answers. Even if we would have the ci system that could do it, it would be quite hard to maintain. Same is true for our unit tests. In order to aim for 100% test coverage we have to keep things simple. 

With the "what kind of package do you want to create"-question, we expose our main use cases up front and can test them easily.

Though, if we come to the conclusion that the all-in-one package use case should be covered by bobtemplates.plone, we have to decide if we want to add "Everything" as an option to the "what type of package would you like to create"-question:

1) Basic
2) Dexterity
3) Theme
4) Everything (Dexterity and Theme)

or if we want to ask those questions individually. Though, I currently don't see how we can test the latter properly (especially if we decide to add even more questions). Think about the logic that is necessary to decide if a question needs to be asked depending on other answers and then try to list all possible outcomes because this is what you need to test.

My recommendation would be to keep the initial type-question as it is and not support the all-in-one use case. If we decide to support the all-in-one use case anyways, we still should still keep the initial question to keep things testable and maintainable.


pbauer:
My opinion differs slightly. 

The most important target audience of bobtemplates.plone are developers (especially beginers) who have a really hard time creating packages with paster etc. But bobtemplates.plone should also be the encouraged best practive for collective- and core-packages. Advanced developers will have no problems removing or adding things to fit their needs. 

For this reason Timo removed questions like profile (if you really don't want a profile you will know why and will know what to remove). The same is true for locales, tests, travis, keyword, license and package-version. It is safe to assume that you either want this or can remove it if you don't. Fewer option also means testing is easier :-)

I think it would suit the normal developers real-life better to not choose between types of packages 

1) Basic
2) Dexterity
3) Theme
4) ...

The one-line UI is also not very good to explain these options. 

I think it would be more accessible if bobtemplates.plone asked if a user wants a theme in their package instead if it wants to create a theme-package. 

I once had to try out all the different package-types of zopeskel and templer to find one that was fitting. Being asked if I want something or not makes more sense to me than having to choose beween options.

To make it as easy as possible I think we should only ask only two questions, both with False as default:

- Add example dexterity content-type and views? [False] 
- Add empty Diazo theme? [False]

This will result in four use-cases we'd have to test:
- Basic Package
- Theme Package
- Basic Package with example content und views 
- Theme Package with example content and views

Timo is the testmaster here but I think we could test these four options (each twice because we also have to test packages with nested namespaces).  


**My 0.02$ on package separation**
The policy/content/theme separation makes less and less sense in my real-life projects. The boundaries between the three is often very opaque. We can have philosophical disputes in our office where a certain piece of code belongs. 

A portlet displaying data from a contenttype belongs in content, the configuration who can add said portlet and where it is visible in policy and the css/js for that portlet in theme. That is insane. 

We turned to having most features in one package called 'myproject.site'. A diazo-theme will mostly end up in a separate package but content and policy becomes site for us. Bigger features that are reuseable in other projects also get their own package (but only with a real usecase, not if we think 'that may be useful sometimes in the future...'. 

I think "it depends" is a valid answer to the question how to structure a project. For beginner developers creating their first projects it will probably not be the global fortune-500 intranet but something smaller. Some content-types, some configuration, some template-overrides and a little light theming. IMHO these kind of projects can and should live in one package. 

For bigger projects you'd probably have a 'suite' or 'site'-package that depends on all the sub-packages (theme, content, workflow, shop, younameit) that are developed for the project often by separate people.

sverbois:
I completely agree with pbauer comment.

jcerjak:
I agree with @pbauer on package separation (+1 on the philosophical disputes in the office :)). We usually also drop the policy package and have a general 'site' package, with a separate diazo theme package. For smaller stuff also everything in one package.

I think that having everything in one package is a perfectly valid use case for simple projects. We should also make it as easy as possible for newcomers to start hacking on Plone and not having to deal with three packages and thinking where something fits. From the Design Goals #29: "Simple is better than complex."

I'd prefer the template generation logic and questions as suggested by @pbauer , if it doesn't complicate the testing too much.

gforcada:
Another way to look at it would be to make the question like:

*Are you developing an add-on for Plone or a Plone website?*

The former will not need a theme nor content types, while the later needs the "all-in-one".

I'm not sure how mr.bob presents questions, but if the explanation can take some *clear and thoughtful* lines, then we are good to go with something like the above I would say and then we reduced the matrix to only two possibilities.

We could also put some ``README.rst`` or top-level comments on certain files to explain what are they for and what else needs to be removed if you want to get rid of it.

tisto:
Last test failure seems to be a regression in Plone 5.0a3: See https://github.com/plone/Products.CMFPlone/issues/381

fulv:
I agree with @pbauer.  I have long been using all-in-one packages, because
the split introduces complexity that in my projects is just useless.  So I
have had to deal with piecing together all-in-one packages from multiple
zopeskel-generated ones, anyway.

I also think @gforcada's idea makes a lot of sense:  add-on, or all-in-one

tisto:
I wasn't aware that the all-in-one approach is so popular. I will try to explain in more detail why I think this is an anti-pattern.

First of all I'm aware that package architecture is hard. I've seen people spreading functionality across packages without thinking a second about package architecture, decoupling or testability. This usually leads to a project with many packages that are so tightly coupled, that they can only be used and tested together anyways.

In this scenario, I agree that many packages do not make any sense at all and it is far easier to put everything in just one single package. Even if you do care about package architecture, you have to put considerable effort into decoupling your components.

Though, tightly coupled packages or one single all-in-one package will make it impossible to properly test your components in isolation. You will end up (or start) with a single "policy" test fixture. All your tests will use that single fixture. Even simple integration tests will always run against your full test setup.

Writing tests in those kind of projects is a nightmare. Your tests will run for ages and your "policy" test fixture is filled with global configuration, that will make it impossible to change things without breaking tests. TDD is just plain impossible with long running tests.

If you don't write tests at all or if you are an integrator with very little experience, having to deal with package architecture might be too much. Though, having different packages for content types, theme and policy at least give you a good start. You can still ruin things if you don't care. Though, telling people that they should give up on package architecture or testability is the wrong message in my opinion.

Our main problem is actually the theming story, because we have just too many ways in Plone to change the appearance of HTML fragments. Therefore we don't know where to put things. We can solve that in the long run by moving to a pure JS frontend and making Plone deliver plain JSON. ;)

In short term we can try to come up with guidlines for developers where to put things. This is what the Plone Training should do.

Bottom line is, I don't see how we can have a Plone training that tells people to put everything into one single package and do proper testing at the same time. One thing that I really like about testing is that if it becomes too hard to test a thing, it usually means you are doing something wrong in your code or package architecture. The right thing to do in those situations is to fix your code, not give up on testing or coding best practices. Being approachable to newbies is a valid goal. Though, I don't see tests as an optional feature but instead as an essential part of software development.

aclark4life:
+1 @pbauer, -1 @tisto (although I appreciate your comments). Let's let people decide later if they want to component-tize. If I am going to use a template generator for Plone I want it to be two things:

- Part of the core (so I know instinctively the core team supports this approach. Even if that approach is dated, I know it was at one time "officially supported".)
- Stupid simple. I don't want to think about policies or themes, etc. I want an "Add on" that customizes Plone to my needs, so I can get my work done.

Remember our target audience is not developers per se, it's "people tasked with getting websites deployed." Plone was most popular when those customizations occurred TTW (and they still do, just look at plone.org:

screenshot 2015-02-22 07 51 20

). If we could build a system that exports TTW customizations into a single add-on package, that would be a game changer. But until then, let's cater to the folks that would most want and use such a system by keeping the template generator "simple":
- Ask me what I want to call it
- Generate boilerplate and/or helpful comments that show me where I can do what inside

do3cc:
+1 @tisto

tisto:
If we want to test bobtemplates.plone against 4 different Plone versions with 3 package types, we end up with a simple 4 x 3 matrix = 12 jobs:

Plone Version x Package Type

This is what I currently aim for. We do not have any overlaps between different options (e.g. theme and dexterity) and we do not support any sopisticated package type options (just the dexterity type name, I'm pretty sure we will have to add some more options in the future).

Before I started to invest time into bobtemplates plone we had two very basic (some might say "alibi") tests that just test the output of the file structure mr.bob generates, nothing more. My approach was to clean up and simplify the package until we have a chance to properly test what we are doing. This pull request is part of the clean up and simplification. I haven't started with writing proper tests, just two high-level acceptance tests that run the tests of the 8 possible outputs of the 4 x 2 matrix mentioned above. We neither have proper unit tests, nor the infrastructure to test against multiple Plone versions (buildout.plonetest is untested and unversioned, we need to find a better solution). This means we are currently only testing a 2 x 2 matrix with some very basic tests (https://travis-ci.org/plone/bobtemplates.plone/builds/51644101).

This is where we are. I would like to go further simplifying things and investing into testing and ci before moving on (this is what this pull request is about). I want it simple and I want 100% unit test coverage.

Here is the alternative:

If we add the all-in-one-package option we already have 16 jobs (4 x 4 matrix). Keep in mind that every single job might fail randomly (buildout anyone?) on travis and that someone needs to take care of this as well.

If we (conceptually) drop the package type option and go with boolean questions for theme/dexterity we have a 4 x 2 x 2 matrix (three dimensions instead of two, same number of jobs of course). If we want to add other options it get's more complex (I'd be open for a discussion if single questions are better for usability, but from a coding/testing point of view the decision should be clear).

If we will have an overlap in functionality between theme/dexterity/whatever things become unmanageable and untestable (something that I already saw happening for the simple questions I removed). The package type option will make sure that we never cross that line because we conceptually separate those "important" questions from the regular ones. It also makes sure we have simple comparisons in the mr.bob code instead of having to combine multiple options.

If we want to add more options our matrix becomes larger and more complex (more options, more dimensions). Not sure how much travis (or we as a community) can handle.

If we want to add package type specific options, testing becomes even more complex (we will have a tree, not a "simple" matrix any longer). I'm not even sure we can properly handle that with the simple design I proposed.

Bottom line is, if we keep things simple, we will have a chance to properly test our code in the future. If we decide to add more complexity before we even test our current code, we will never have a properly tested package. Maybe we should have started from the scratch in the first place.

If anybody who advocates the all-in-one-package solution steps up now and commits themselves to implement the tests and ci infrastructure that is needed, please go ahead. If this does not happen, we have to make a decision what is more important for us right now, proper testing or functionality.

tisto:
Ready for review/merge.

bobtemplates/plone_addon/src/+package.namespace+/+package.name+/tests/test_+package.dexterity_type_name_lower+.py.bob
```diff
@@ -0,0 +1,44 @@
...
+from zope.component import queryUtility
+from zope.component import createObject
+from plone.app.testing import setRoles
+from plone.dexterity.interfaces import IDexterityFTI
+from plone import api
+
+from {{{ package.dottedname }}}.testing import {{{package.uppercasename}}}_INTEGRATION_TESTING  # noqa
+from {{{ package.dottedname }}}.interfaces import I{{{ package.dexterity_type_name }}}
+
+import unittest2 as unittest
```
do3cc: unittest2 is only needed for python 2.6

do3cc:
@tisto, is it ok for you if we only merge after our training which end friday next week? I can help out with rebasing

bobtemplates/plone_addon/src/+package.namespace+/+package.name+/tests/test_+package.dexterity_type_name_lower+.py.bob
tisto: Thx for the hint. Will fix that.

tisto:
@do3cc sure. do you plan to make a release before the training?

do3cc:
Probably not

pbauer:
I created a pull-request https://github.com/plone/bobtemplates.plone/pull/51 that automatically builds and tests all possible combinations and is dead-simple to update when something changes. I'm willing to maintain that as long as someone else helps me with setting up the ci-infrastructure. 

I still belive very strongly that we should go with the option that is most accessible, which means having multiple questions about features not one about a package-types and thus enabling people to create a all-in-one package. The default-answers and the documentation should at least discourage having a diazo-theme in the same package as the rest. 

Sorry for being brief but I really have to work on the plone5-training.

smcmahon:
A late entry.

I'm with tisto on this one. jcerjak cited the design goal: "Simple is better than complex." Let me trump that with the zen of Python:  "Namespaces are one honking great idea -- let's do more of those!"

My chief interest in a skeleton generator is for the sake of new developers. When I teach, I am not teaching them theming and dexterity at the same time. There are indeed overlapping concepts, but I'm trying to get them into the "dexterity namespace" or the "plone.app.theming namespace". The more clearly designed for a purpose a skeleton package is, the easier it is to get them going without fear of breaking stuff.

Once we've been through Dexterity and theming, it's easy to step back and say that if you want to do it there's no reason not to pull these things together.

And, tests running faster ... that sure helps teaching testing. Tests running slower, we lose them.

djowett:
FWIW & arriving late to the discussion, I think asking about "type of package" is the wrong question, and *more importantly*, the *wrong* **time** to ask it to a newbie. 
Often they just want a package & to add Dexterity (or theme) to it later.  They may not even know what Dexterity is when mrbob asks the question.  I would say the questions proposed by @gforcada or @pbauer are better.

However, I would say that it would be better & less load on Travis to have a separate bobtemplates.dexterity (as mentioned in #35), which relies on the generated egg from this template & also makes checks that everything to get dexterity types working is in place or if not *added in* to the generated egg.  

So load on Travis would be:

- bobtemplates.plone  4
- bobtemplates.dexterity 4
- bobtemplates.theme 4
Total 12, and *not* increasing logarithmically.  Also - since they are in different packages - they only need run when that package is changed (Ok, I'm simplifying but not much).

That would however mean that placeholders have to be left in the generated egg for bobtemplates.dexterity & similar to use to insert text.  I know that's a throwback to Zopeskel / paster (Remember `# -*- Extra requirements: -*-`), but was that really the bad bit?

Ok, that's my piece. I'm aware it may be too late, so not to worry - I'm sure we'll get something useful whatever & it doesn't stop anyone making their own bobtemplate. 

Cheers,
Daniel

gforcada:
IMHO trying to bolt everything on a single package is a bit of nonsense... if we do really decide to have different types of packages then let's make one repository per package, so testing doesn't get so much of a problem and each package can evolve independently.

Bonus points on keeping them sane and fairly equal so that porting fixes from one to another is feasible, although that (for me) would be a bonus not a requirement...

tisto:
@djowett the type question has a default, so newbies can just ignore it, if they don't understand it. I just gave a training with the latest version of bobtemplates.plone and people had problems with basically all questions. This is why I just removed most of them, this makes a real difference I think. In the end having one question (with a default) that you don't understand is better than having two questions you don't understand.

@djowett @gforcada mr.bob currently does not allow to build upon another template. This is why I went with the all-in-one approach. Separating them would mean duplicating lots of code, which would be the worst option in my opinion. As long as we keep things simple, I guess we can deal with that (for now).

My pull request is ready, all three options are now tested with basic acceptance tests and pass on travis. Next step would be to further clean up things and write unit tests. I might have time to put some more effort into the theme part in the next weeks, because I have to prepare another training.

djowett:
@tisto thanks for replying.  Point conceded.... You are right in saying that mr.bob doesn't make building on another template easy (+ developers could have changed the code between running one template & the other) - so it's not worth bending over backwards to make something more difficult to support ;-)
It's probably better for people's understanding of dexterity to make their second type manually (even if it's by copy & paste).
And great to hear you are so busy training new Ploners

hvelarde:
2 years ago we started this: https://github.com/plonegovbr/bobtemplates.plonegovbr

it may have some additional stuff or ideas for this package.

do3cc:
obrigado! Eu vou olhar se nos podemos usar alguma coisa.

jensens:
just a very late input:
we here started to build one package for a customer project and then use namespaces inside adressing the different domains, example:
package/egg: kup.site.mycustomer
subfolder: kup.site.mycustomer.theme, kup.site.mycustomer.content, kup.site.mycustomer.workflow, ...
this results in several profiles (for each subfolder where its needed).
I just wonder if a base profile and then adding domains to it later would be a more practical approach.

Anyway, with current mr.bob this is probably not possible, since this would need tool that reads the current code and adds stuff (agx on a different level does this btw.).

but just my 0.02 cents. this is not very helpful for this actual issue problem i guess.

jensens:
fun fun fun: and after reading all the threads it really funny to find here two different patterns how the involved developers think (and this can be expanded to other discussions):

- "design thinking" (user needs first, technology has to follow)
- technical first (technical approach, how to make it follow best practices in testing, simplify to technicla needs, architecture, user has to follow this)

both has its pros and cons. and both is needed. if this analysis and synthesis is combined well its the best we can get.

do3cc:
After finishing the training, I have a high confidence that we can provide simple mr.bob template with a documentation providing links to up to date and tested code snippets that show how to implemented missing features together with tests

gforcada:
@do3cc that's really cool to hear!

tisto:
Will merge this pull request now. @pbauer @do3cc and I discussed the issues mentioned above last week and we agreed to start simple and discuss the all-in-one package use case at a later point, when we have a stable and well-tested foundation.

PR #49, opened 2015-02-23
WIP idea for extending

do3cc:
When one creates a package, it is not always clear, what code is needed.
This PR proposes the idea of providing a bunch of example code that can be copied/appended to the existing package.
they reside in a directory that is not imported in any way and can be excluded in Metadata.in so that it also won't exist as dead code in eggs.
What is still missing is a simple shell script or make file that can copy new files and append content to existing files. Appended files still need clean up. Which files need clean up a code-analysis run would show (Except for zcml... :disappointed:)

This is open for discussion, if you don't like it, please provide ideas for alternatives.
I am looking for templates to include

- a viewlet
- a portlet
- a view
- a css file
- a js file
- a contenrule
- a behavior
- an event handler
- a browser view to be used as conditions in tal statements (for workflows, for example)
- memberdata extensions
- custom indexer
- placeful workflows
- registry and control panel views
- new roles

I think these are common use cases. Common to those use cases is that you won't know of all of them that you need them at the start of the project. Common to most of them is that you might not be aware of the current best practices on how to integrate AND test them.

### DIFF
--- a/bobtemplates/plone_addon/src/+package.namespace+/+package.name+/more_templates/contenttype/interfaces.py.bob
+++ b/bobtemplates/plone_addon/src/+package.namespace+/+package.name+/more_templates/contenttype/interfaces.py.bob
@@ -1,0 +1,14 @@
+from zope import schema
+from zope.interface import Interface
+from {{{ package.dottedname }}} import _
+
+class IThing(Interface):
+    title = schema.TextLine(
+        title=_(u"Title"),
+        required=True,
+    )
+
+    description = schema.Text(
+        title=_(u"Description"),
+        required=False,
+    )
--- a/bobtemplates/plone_addon/src/+package.namespace+/+package.name+/more_templates/contenttype/profiles/default/types.xml
+++ b/bobtemplates/plone_addon/src/+package.namespace+/+package.name+/more_templates/contenttype/profiles/default/types.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0"?>
+<object name="portal_types" meta_type="Plone Types Tool">
+  <!-- For each content type, you create a new files types/{{{ name }}}.xml
+       Then you create a new entry below, where the attribute name must match
+       {{{ name }}} from above. The other other attribute you will only modify
+       if you do something non dexterity like -->
+  <object name="thing" meta_type="Dexterity FTI"/>
+</object>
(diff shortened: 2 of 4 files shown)


aclark4life:
Nice

gforcada:
Wouldn't make sense to show those snippets on ``bobtemplates.plone`` documentation? Just like you can go to http://patternslib.com/#documentation and get and overview and examples about what each pattern does. The same could be done with ``bobtemplates.plone`` docs.

This way they can freely copy and paste the code, but most importantly, they have some short documentation about what they are about to copy and references to docs.plone.org for further reading...

If we really target beginners, then a list showing to them "portlet" next to "viewlet" will be probably confusing enough.

*/me starts to wonder if bobtemplates.plone should be really a fancy web form that generates a zip file to be downloaded*

do3cc:
@gforcada I like your idea much more than what I proposed here. But I am not sure how I could test it then.

tisto:
When I gave a plone testing training some weeks ago I created a package with example tests and the implementation to make them pass. My training were basically sphinx docs that included those test snippets. 

This way I could make sure the documentation is up to date and the code is tested at the same time. People could even go ahead and look at the real implementation. 

Though, I'm not sure though if bobtemplates.plone is the right place for this. I would rather put that into the regular plone docs. If we ship all this code people will keep it in their repo. It will become outdated. Rember "This product may contain traces of nuts"?

do3cc:
I never got this

gforcada:
When I was saying shpinx docs I was talking about sphinx docs from bobtemplates.plone *itself* rather than what package gets generated with bobtemplates.plone

do3cc:
I am agnostic as to where the code is. All I want is that is is easily reachable and easy to keep in sync in terms of best practices.

gforcada:
So bobtemplates.plone should generate some stub documentation (IMHO) for each package, but also have its own documentation, put it online on docs.plone.org with vesions and so on, where we give all this snippets.

Actually the package generated docs could already point to that. i.e.

In ``docs/source/develop.rst``:
To add more functionality on this package, please refer to http://docs.plone.org/XXX/YYY/bobtemplates.plone/extend.rst.

What about this?

tisto:
Ok, I guess we should discuss with the docs team (cc @svx @polyester) where to put this kind of code.

Here is my plone training code:

https://github.com/tisto/plonetraining.testing
http://plonetrainingtesting.readthedocs.org/en/latest/index.html

Part 4-6 are not tested yet. My idea was to put this directly into the plone.app.testing docs as soon as I find time.

jensens:
whats the state here? close?

do3cc:
yes

PR #51, opened 2015-02-26
Automatically build and test all possible combinations of packages

pbauer:
Do not merge yet. This is a proof of concept that it is possible to build and tests all possible combinations of packages. One problem is that running the tests will take quite some time (especially if we add more options) and travis with the limit of 10 minutes will not be enough.

### DIFF
--- a/test_all_combinations.py
+++ b/test_all_combinations.py
@@ -44,12 +44,6 @@ def combine(list_a, combinations):
     return retval
 
 
-def is_plone5(version):
-    if version.startswith('5'):
-        return True
-    return False
-
-
 def generate_all_combinations():
     keys = SOURCE.keys()
     first = SOURCE[keys[0]]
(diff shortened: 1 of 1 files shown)


test_all_combinations.py
```diff
@@ -0,0 +1,135 @@
...
+    return False
+
+
+def generate_all_combinations():
+    keys = SOURCE.keys()
+    first = SOURCE[keys[0]]
+    combinations = [[x] for x in first]
+    for key in keys[1:]:
+        combinations = combine(SOURCE[key], combinations)
+    combinations = [dict(zip(keys, x)) for x in combinations]
```
do3cc: I thought you don't like functional methos like zip and map and so on

pbauer: I like them when you write them for me.

tisto:
@pbauer if you think that being able to run all those combinations is the main problem, you entirely missed the point I was trying to make in #46.

tisto:
https://en.wikipedia.org/wiki/Unit_testing

do3cc:
I don't know this guy

PR #54, opened 2015-03-18
use PloneWithPackageLayer, testing.zcml and testing profile

gotcha:
Setting up fixture this way  enables to test the fixture TTW
with a zope instance started with testing.zcml loaded.

This does not impact the standard use of the add-on.

### DIFF
--- a/bobtemplates/plone_addon/buildout.cfg.bob
+++ b/bobtemplates/plone_addon/buildout.cfg.bob
@@ -16,6 +16,7 @@ eggs =
     Plone
     Pillow
     {{{ package.dottedname }}} [test]
+zcml = {{{ package.dottedname }}}-testing
 
 
 [code-analysis]
--- a/bobtemplates/plone_addon/src/+package.namespace+/+package.name+/profiles/testing/metadata.xml.bob
+++ b/bobtemplates/plone_addon/src/+package.namespace+/+package.name+/profiles/testing/metadata.xml.bob
@@ -1,0 +1,7 @@
+<?xml version="1.0"?>
+<metadata>
+  <version>1000</version>
+  <dependencies>
+    <dependency>profile-{{{ package.dottedname }}}:default</dependency>
+  </dependencies>
+</metadata>
--- a/bobtemplates/plone_addon/src/+package.namespace+/+package.name+/setuphandlers.py.bob
+++ b/bobtemplates/plone_addon/src/+package.namespace+/+package.name+/setuphandlers.py.bob
@@ -1,8 +1,9 @@
 # -*- coding: utf-8 -*-
 
 
-def isNotCurrentProfile(context):
-    return context.readDataFile('{{{ package.longname }}}_marker.txt') is None
+def isNotCurrentProfile(context, marker='marker'):
+    filename = '{{{ package.longname }}}_{}.txt'.format(marker)
+    return context.readDataFile(filename) is None
 
 
 def post_install(context):
@@ -10,3 +11,10 @@ def post_install(context):
     if isNotCurrentProfile(context):
         return
     # Do something during the installation of this package
+
+
+def test_fixture(context):
+    """Test fixture setup"""
+    if isNotCurrentProfile(context, 'testing'):
+        return
+    # Do something for test fixture
--- a/bobtemplates/plone_addon/src/+package.namespace+/+package.name+/testing.py.bob
+++ b/bobtemplates/plone_addon/src/+package.namespace+/+package.name+/testing.py.bob
@@ -3,40 +3,24 @@
 from plone.app.contenttypes.testing import PLONE_APP_CONTENTTYPES_FIXTURE
 {{% endif %}}
 from plone.app.robotframework.testing import REMOTE_LIBRARY_BUNDLE_FIXTURE
-from plone.app.testing import applyProfile
 from plone.app.testing import FunctionalTesting
 from plone.app.testing import IntegrationTesting
-{{% if not plone.is_plone5 %}}
-from plone.app.testing import PLONE_FIXTURE
-{{% endif %}}
-from plone.app.testing import PloneSandboxLayer
+from plone.app.testing import PloneWithPackageLayer
 from plone.testing import z2
-from zope.configuration import xmlconfig
 
 import {{{ package.dottedname }}}
 
 
-class {{{ package.browserlayer }}}(PloneSandboxLayer):
-
+{{{package.uppercasename}}}_FIXTURE = PloneWithPackageLayer(
 {{% if plone.is_plone5 %}}
-    defaultBases = (PLONE_APP_CONTENTTYPES_FIXTURE,)
+    bases=(PLONE_APP_CONTENTTYPES_FIXTURE, ),
 {{% endif %}}
-{{% if not plone.is_plone5 %}}
-    defaultBases = (PLONE_FIXTURE,)
-{{% endif %}}
-
-    def setUpZope(self, app, configurationContext):
-        xmlconfig.file(
-            'configure.zcml',
-            {{{ package.dottedname }}},
-            context=configurationContext
-        )
-
-    def setUpPloneSite(self, portal):
-        applyProfile(portal, '{{{ package.dottedname }}}:default')
-
-
-{{{package.uppercasename}}}_FIXTURE = {{{ package.browserlayer }}}()
+    zcml_package={{{ package.dottedname }}},
+    zcml_filename='testing.zcml',
+    gs_profile_id='{{{ package.dottedname }}}:testing',
+    name='{{{ package.browserlayer }}}',
+    additional_z2_products=()
+)
 
 
 {{{package.uppercasename}}}_INTEGRATION_TESTING = IntegrationTesting(
--- a/bobtemplates/plone_addon/src/+package.namespace+/+package.name+/testing.zcml.bob
+++ b/bobtemplates/plone_addon/src/+package.namespace+/+package.name+/testing.zcml.bob
@@ -1,0 +1,25 @@
+<configure
+    xmlns="http://namespaces.zope.org/zope"
+    xmlns:five="http://namespaces.zope.org/five"
+    xmlns:genericsetup="http://namespaces.zope.org/genericsetup"
+    xmlns:i18n="http://namespaces.zope.org/i18n"
+    i18n_domain="{{{ package.dottedname }}}">
+
+    <include file="configure.zcml" />
+
+    <genericsetup:registerProfile
+        name="testing"
+        title="{{{ package.dottedname }}} tests"
+        directory="profiles/testing"
+        description="Installs the tests fixture for {{{ package.dottedname }}} add-on."
+        provides="Products.GenericSetup.interfaces.EXTENSION"
+        />
+    
+    <genericsetup:importStep
+        name="{{{ package.dottedname }}}-testFixture"
+        title="{{{ package.dottedname }}} test_fixture import step"
+        description="Test fixture import step from {{{ package.dottedname }}}"
+        handler=".setuphandlers.test_fixture">
+    </genericsetup:importStep>
+
+</configure>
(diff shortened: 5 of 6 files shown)


gforcada:
There were some arguments about not adding (if really really not needed) a testing profile. Basically due to adding global state on all tests, which make them more fragile and not as self contained as they should be.

@tisto most probably will give you some other reasons.

btw. as the branch is still not merged, could you squash together all this pep8 commits? ``git rebase -i master`` is your friend.

tisto:
@gotcha do you remember that we had the same discussion about the templer long time ago? :)

I still believe that test profiles are wrong. They introduce global state to the test setup, which makes it harder for people to understand what goes on in a test. Since tests are documentation for developers that's a pretty bad thing in my opinion. A global test setup encourages people to add more stuff to the global state which makes the entire test setup brittle. If you want to change something in your global test setup there is a very good chance that this will break quite a few tests.

gotcha:
I totally agree with keeping the global state as small as possible.
Nevertheless, what is the difference between doing stuff in a layer setup method and having the same method call a generic setup test profile ? As long as both are kept as lean as possible...
The latter enables checking what is done by the layer ttw.

gotcha:
@gforcada Squash done. Thanks of reminding me.

tisto:
Well, we get more files and more boilerplate code. One single testing.py, where people can look up the entire test setup, is far more readable than a configuration that is spread across different files. 

I know you don't like that argument, but we are also changing the way we taught people testing for quite a while. We would have to change our docs, our trainings and our books. In general, change is perfectly ok, if there is a strong argument for it. Though, I don't see that here.

What do you mean by "checking ttw"?

gotcha:
If an instance is started with testing.zcml loaded, the test profile can be used to instantiate a Plone site. This way, the developer can also check TTW that his setup does what he wants.

Now, I really dislike the argument about existing docs. If I take this type of reasoning to its extremes, we should not use plone.app.contenttypes as it forces us to change our documentation. Or less extreme, we should keep using PloneTestCase for the same reason. ;-)

tisto:
I see that a ttw check could be a nice feature for newbies. Though, I'm not sure it that really helps that much in real life. Firing up a real instance and applying a test profile ttw and debugging a test setup are still two different things. If I want to debug test setups, it still could be that an error only occurs on the test and not on the profile. In additon, if we want people to use that feature, we force them to put more stuff into the global state to be able to use it (something that we agree we both don't want to encourage).

I agree that "we will have to change to docs" is a weak argument, if we are changing something for the better. I just think that we have to carefully weight pros and cons for every possible change. In this particular case my opinion is still that the cons outweigh the pros.

gotcha:
How does adding a testing profile that is only added by the testing zcml adds to the global state ?

tisto:
Maybe I got this wrong. If I want to be able to test something in my test setup ttw (by starting an instance and applying the testing profile manually), I have to add it to the global state, otherwise it will not show up in my manual test, right? Therefore it is even more tempting for newbies to add stuff to the global state in order to be able to test it manually.

gotcha:
I think you need to define "the global state".

I have the feeling we are back to my words that test layer setup should be as small as possible because it is global.

And you did not answer to my question where I asked what the difference is between code called by the layer rather than by the testing profile.

The main goal of this pull request is about using PloneWithPackageLayer which hides a lot of complexity of plone.app.testing by making the developer reuse concepts that he nevertheless needs to understand like registering stuff through ZCML, GS and import steps.

gotcha:
When I search Github code, I find almost two hundred occurences of PloneWithPackageLayer. IOW, other than me think it helps them.

frisi:
personally, i'm using `PloneWithPackageLayer` because i need not use that much boilerplate code such as registering the `zope.configuration.xmlconfig` or call `z2.installProduct` in `setUpZope` and do not need to remember to uninstall in `tearDownZope`

i think we can agree that there is nothing bad about PloneWithPackageLayer, right?

if i understand your discussion right - it's mainly about having an additional configuration file for tests (testing.zcml) and if this should contain another GS profile that does the test-setup.

personally i prefer to do the test setup in `TestCase.setUp`. somethimes this is also calling python methods shared with GS importsteps or running GS steps - why not?

could we agree on not adding a `testing.zcml` in the bobtemplate but load the package's `configure.zcml` (and maybe also apply the profile) using `PloneWithPackageLayer`?

@gotcha inspecting a test-setup TTW can really be helpful - i agree.
you can do that without adding another profile, too: add the method below to your TestCase, add a pdb breakpoint in your test_method and call `self._access_server`:

```python
def _access_server(self):
        from Testing.ZopeTestCase.utils import startZServer
        host, port = startZServer()
        return 'http://%s:%d%s' % (host, port, '/'.join(self.portal.getPhysicalPath()))
``` 

regards, fRiSi

gotcha:
@do3cc @djay @tomgross @bloodbare @simahawk @miohtama 

Would you mind giving your opinion on this ?

tisto:
Before putting more effort into this discussion. @gotcha, you are proposing to include a method into a "best practice" package that is completely undocumented:

https://pypi.python.org/pypi/plone.app.testing/5.0b2

There is just a small note in the changelog, nothing more.

Even if I would agree with your arguments (which I don't), there is no way that we can add a feature that has no documentation at all.

gotcha:
@frisi I'd be happy with using PloneWithPackageLayer to include configure.zcml. However, that does not hint how to add test setup that needs to happen at layer setup time.

@tisto You are right that PloneWithPackageLayer should be documented in plone.app.testing README.

gotcha:
https://github.com/plone/plone.app.testing/pull/11

tisto:
@frisi I'm afraid we can not agree on that. We have a well defined, well documented way of doing tests in Plone. When I do testing trainings, I can explain people how plone.app.testing works, with just going through testing.py. It is relatively simple and it is clean. It does not hide too much and it is not too much boilerplate code either. I can tell them they can look things up later in docs.plone.org and in PPD and it will all be the same stuff.

We would loose all that, if we would define a new testing.py best practice. Even if we would do it right, we would have to rewrite everything. Otherwise we will end up with two different ways of doing the same thing, which is much more confusing for people than a few more lines of code (which actually help people to understand how things work).

I put a lot of effort into cleaning up the test setup, make it follow best practices, and fixing test isolation problems in coredev in the last two years. I will not amend all coredev tests to a new best practice, that does not have a considerable advantage (not to speak about the disadvantages I pointed out).

Please ask yourself. Are you willing to do the necessary work yourself? Do you really think it is worth the effort? It is easy to propose a change. Cleaning up the consequences is not...

gotcha:
Why does including PloneWithPackageLayer as a best practice for add-on developers would force rewriting all Plone core tests setup ? This is totally unrelated imo. 

Why would we vote to show and teach complexity to devs when simplicity is enough ?
And yet it is very important that when complexity is needed it is documented.

tisto:
@gotcha because otherwise we will end up with two different ways of doing things, which is even worse (see grok/zcml, zca component registration, etc.). My experience is that people do not have a problem with a few more lines if things are well documented and straight forward (and if that helps them to understand the underlying system). They do have a big problem if you tell them that there are two ways of doing the same thing and that they have to decide for themselves (at least at some point).

frisi:
@tisto i was just trying to bring this discussion to an agreement - and failed ;-)

if using a utility layer like `PloneWithPackageLayer` in bobtemplate.plone means it's considered best practice which on the other hand means all plone.* packages should follow this best practice then this change would mean a lot of consequences/work - true.

I'll refrain from debating whether these implications really have to be the way you outlined.
I can live with and without PloneWithPackageLayer

tisto:
@frisi sure, I got that. You probably don't know that @gotcha and I had the exact same discussion long time ago with templer. You just got in the middle. I think the only way to solve this is, that @gotcha and I meet in person and discuss this over a drink. ;)

do3cc:
I am -1 on this.

- setup.py will become a location to modify state just for tests. Bad
- test profiles, err, please no. These profiles always show up in plone and will add confusion there. People might not be sure which profile needs modification. The benefits you describe do not outweigh the disadvantages
- The PloneWithPackageLayer adds a wrapper to hide things. testing.py makes it quite explicit when what gets loaded. This aids in understanding. Additionally nobody spends a lot of time in testing.py. Shaving of a few lines by a helper does not do much in terms of saving developer time.

In my opinion, these changes are not an improvement. So the question if these changes are good enough to warrant updating our training material, does not apply.

gotcha:
@do3cc 

- I do not understand what you mean with setup.py. Can you expand ?
- Test profiles are registered by testing.zcml. Iow, they do not show up anywhere unless explicitly loaded.
- I guess we should also avoid using plone.api as it hide things. And it is not about saving lines. But about postponing learning plone.testing API until really needed.

plone-foundation:
Can someone sum up this thread for me? Godefroid wants to merge something
that Timo does not consider best practice, is that correct? In lieu of
cocktails (which may be required to make progress), I'm interested in
learning more about Plone's current testing "best practices" via this
thread (as it has obviously struck a nerve.)

On Thu, Mar 19, 2015 at 3:01 PM, Godefroid Chapelle <
notifications@github.com> wrote:

> @do3cc <https://github.com/do3cc>
>
>    - I do not understand what you mean with setup.py. Can you expand ?
>    - Test profiles are registered by testing.zcml. Iow, they do not show
>    up anywhere unless explicitly loaded.
>    - I guess we should also avoid using plone.api as it hide things. And
>    it is not about saving lines. But about postponing learning plone.testing
>    API until really needed.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/plone/bobtemplates.plone/pull/54#issuecomment-83718963>
> .
>



-- 
Alex Clark · http://about.me/alex.clark

aclark4life:
Can someone sum up this thread for me? Godefroid wants to merge something that Timo does not consider best practice, is that correct? In lieu of cocktails (which may be required to make progress), I'm interested in learning more about Plone's current testing "best practices" via this thread (as it has obviously struck a nerve.)

do3cc:
@gotcha
- Sorry, I meant setuphandlers.py, here: https://github.com/plone/bobtemplates.plone/pull/54/files#diff-28120dab44740c0624cf154fdeb5a7c7R16 I now have an additional point to set up my test layer.
- You are right, I did not notice that the layer gets loaded in testing.zcml
- I am not against using plone.api. The difference between the helper you propose to use here and the helpers from plone.api is that I need to use plone.api in my day to day business. Testlayers otoh I barely need to touch.
I can see a distinct advantage when seeing side by side how to modify a workflow state with and without plone.api.
This advantage is much smaller with the helper method. It reduces the number of methods and imports by introducing a single method with a big call signature. Another difference to the comparison to plone.api, I don't have to memorize the right import locations and the right call order for testing.py. The code already is there, nobody needs to memorize this.

Finally, I'd like to separate two things. 
Independently from the helper method, I am against providing the explicit test infrastructure to simulate the test environment with a real Plone site. You mentioned it as testing live.
I see the benefit of providing separate zcml configurations. I can, for example, provide a Viewlet but not register it for a specific viewlet manager, so that the integrator can choose where to add the viewlet, or I provide different workflow defaults for new permissions I provide. I do this kind of thing in our Mastering Plone training: https://github.com/plone/bobtemplates.plone/pull/54/files#diff-28120dab44740c0624cf154fdeb5a7c7R16
These kind of things I'd not tie to a testing.zcml, I'd call it a demo.zcml. Or did I misunderstand your idea for testing.zcml here?.
Also I'd keep it out of this package. This package is fairly basic so far and we should stick to this for now. I don't know if you've read the discussion in #46 but the idea currently is to provide a documentation with ready made copy&paste snippets of examples. How to add browser views, how to add viewlets etc. Putting a demo.zcml in there would be better, at least for now.

tisto:
@gotcha I will close this pull request now, are you ok with that? I'm happy to discuss this in person next time we meet or during a hangout. Though, I don't think we will be able to reach an agreement via github messages. Feel free to re-open or comment if you disagree.

gotcha:
I think I might spend some time extending mr.bob so that I can extend other templates. In order to avoid this type of long political discussions.

tisto:
Sure! That would be a nice addition. Do you plan to go to https://plone.org/events/community/anniversary-sprint-2015 or europython this year? Would be nice to meet and discuss things...

gotcha:
I still need to decide for anniversary. Not possible for Europython.

PR #62, opened 2015-05-28
don't add dx-type interface unless needed

pbauer:


tisto:
@pbauer please add a changelog entry.

bobtemplates/plone_addon/src/+package.namespace+/+package.name+/interfaces.py.bob
```diff
@@ -2,13 +2,16 @@
 """Module where all interfaces, events and exceptions live."""
 
 from {{{ package.dottedname }}} import _
```
tisto: @pbauer this leads to a code analysis violation since the _ import is unused. Not sure why travis did not complained about it.

pbauer: fixed

pbauer: Ah, fun. You fixed it as well :-)

Issue #28, opened 2015-02-09
Add code-analysis to travis and fail the build on violations.


tisto:
I usually run code-analysis with plone.recipe.codeanalysis on travis as a first step. This way the build fails as early as possible on pep8 violations etc.. This might sound strict but 100% compliance with code-analysis is the only way to make it work in my experience. People can always remove that if they don't agree but I would consider that our best practice. Opinions?

Here is an example implementation:

https://github.com/plone/plone.restapi/blob/master/.travis.yml
https://github.com/plone/plone.restapi/blob/master/travis.cfg#L19

pbauer:
+1

gforcada:
:+1:

cedricmessiant:
I agree too

gil-cano:
+1

gforcada:
That's already working like this for quite a long time.

PR #67, opened 2015-06-18
Enable code analysis on travis

gforcada:
And fail if it does not pass.

Fixes: https://github.com/plone/bobtemplates.plone/issues/28

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,8 +4,8 @@ Changelog
 0.11 (unreleased)
 -----------------
 
-- Nothing changed yet.
-
+- Enable code analysis on travis and fail if the code does not pass.
+  [gforcada]
 
 0.10 (2015-06-15)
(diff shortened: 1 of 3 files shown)


bobtemplates/plone_addon/buildout.cfg.bob
```diff
@@ -28,6 +28,8 @@ flake8-extensions =
     flake8-blind-except
     flake8-debugger
     flake8-coding
+return-status-codes = True
```
tisto: This should go into travis.cfg IMHO. We do not want to return status codes locally.

gforcada: Aha, I did not notice that there was a travis.cfg, I will fix it right away

tisto:
@gforcada Changelog entry please :)

PR #68, opened 2015-06-20
Fix coverage on travis template. Fixes #58

### LINKED ISSUE #58: Section 'createcoverage' not defined.
travis.cfg.bob refers a createcoverage section, which is not defined.
.travis.yml.bob also include a bin/createcoverage call

gil-cano:


gforcada:
@gil-cano I guess you want to change the toplevel .travis.yml to run coverage as well?

gil-cano:
@gforcada actually I just want to fix the travis.cfg file created in new packages since travis fails with:

    While:
      Installing.
      Getting section createcoverage.
    Error: The referenced section, 'createcoverage', was not defined.

PR #70, opened 2015-07-01
Relax coverage checking on namespace's __init__.py

djowett:
just an idea... - but helps if running coverage

I just copied what plone.api does
https://github.com/plone/plone.api/blob/master/src/plone/__init__.py

### DIFF
--- a/bobtemplates/plone_addon/src/+package.namespace+/__init__.py
+++ b/bobtemplates/plone_addon/src/+package.namespace+/__init__.py
@@ -1,7 +1,2 @@
 # -*- coding: utf-8 -*-
-# See http://peak.telecommunity.com/DevCenter/setuptools#namespace-packages
-try:
-    __import__('pkg_resources').declare_namespace(__name__)
-except ImportError:
-    from pkgutil import extend_path
-    __path__ = extend_path(__path__, __name__)
+__import__('pkg_resources').declare_namespace(__name__)


gforcada:
We could actually change that to a simple:

```
# -*- coding: utf-8 -*-
__import__('pkg_resources').declare_namespace(__name__)
```

I would assume that nowadays, and specially with newer distributions made out of bobtemplates.plone, one can expect setuptools to be around...

djowett:
Ah, I did wonder, but wasn't sure of the history of this....
If there is no setuptools, then setup.py will fail right?

gforcada:
I'm not 100% sure what would be the behavior, although most probably other things will fail as well if setuptools is not around either :)

Reading http://peak.telecommunity.com/DevCenter/setuptools#namespace-packages it seems that the timeframe is around setuptools 0.6/0.7 so anything newer is probably safe without the try/except, but I have no real proof of that. Any other opinion on this?

djowett:
@gforcada I think you're reasoning is sound, but I would prefer others chipped in to confirm.

tisto:
If setuptools 0.6 will fail we can not remove that try/except. Plone 4.x still uses setuptools 0.6. We have to support at least Plone 4.x.

djowett:
Good point @tisto. I'm scrapping this PR & reverting to the original commit at #71 (to keep the history clean).  Is that safe?

thet:
the ``declare_namespace`` method is available at least since setuptools 0.6b1 - i checked that by downloading and grepping: https://pypi.python.org/packages/source/s/setuptools/setuptools-0.6b1.zip#md5=000447ea5cf11ebaecc4f6ce9812a561

i think, it would be safe to simplify as suggested by @gforcada 

btw., here is an explanation about the difference between ``declare_namespace`` and ``extend_path``. i'm wondering, if this means, that with ``extend_path`` other packages cannot add something to an already declared nested namespace?

djowett:
Ok, based on @thet's comment I'm reopening, obviously only one of #70 or #71 should be merged, (and not by me) but we need some consensus.   @tisto can you comment on the above statement by @thet?
Anyone else want to give their 2c?

jensens:
Btw.: In addons managed by me I removed the try except everywhere and never had any problems.

tisto:
@jensens removing try/except will work even for Plone 3 projects?

thet:
I think he meant that only this is left in ``__init__.py``:
```
__import__('pkg_resources').declare_namespace(__name__)
```
A nice oneliner. All of our packages are depending on setuptools anyways, so this should actually be safe.

jensens:
yes, what @thet said. This works fine with setuptools 0.6.c11 used by Plone 3.3

jaroel:
We stopped supporting non-pkg_resources a long long time ago:
At least 2011, probably before that.
https://github.com/plone/plone.app.collection/blob/9d937bacdd57e9d1b92dc02ca39a63d2f256132b/plone/__init__.py

jaroel:
@djowett could you rebase this so it is one commit?

djowett:
@jaroel sorry I won't get it done today, but isn't it good to have 2 commits to show the motivation & discussion anyway?

jaroel:
There's really not that much to discuss :)
We haven't support non-setuptools environments since at least 2011.

jaroel:
Done :)

PR #71, opened 2015-07-02
Relax coverage checking on namespace's __init__.py

djowett:
New branch to replace #70, which got mangled.

This helps if running coverage on the generated egg

See [what plone.api does](https://github.com/plone/plone.api/blob/master/src/plone/__init__.py)


jaroel:
Dup #70

PR #75, opened 2015-07-14
Add uninstall-profile

pbauer:
This is for Plone 5 only since the uninstall-profile is only automatically run on uninstalling with Products.CMFQuickInstallerTool >= 3.0.9 (Plone 5.0b3 and above).


jensens:
:+1:

PR #77, opened 2015-07-24
Plone 5 already depends on p.a.dexterity and p.a.theming

pbauer:
@tisto Right?


pbauer:
@tisto can you make sense of the failures?

tisto:
Seems like one of the profiles is not already present in Plone 5. I'm not sure if we should really do this. I tried to remove as many if/else as possible when I did the cleanup, because they can make it really hard to figure out what goes on. Especially when we start to combine if/else with multiple statements like in this pr. Personally I would prefer to keep the profile loads to keep things simple, even though they might not be necessary any longer in Plone 5.

pbauer:
Agreed.

Issue #81, opened 2015-08-08
Does not deal well when creating Dexterity content types with two words


pigeonflight:
How to replicate this issue.

Create a new package using mr.bob, when asked select "Dexterity" then name the content type something like "Car Part".

After adding it to you buildout and relaunching Plone you'll get an error like this:

     SyntaxError: invalid syntax (interfaces.py, line 14)

You will see that interfaces.py contains a class with a space in it (ICar Part) in this case, which is not valid for Python.

davilima6:
If content type is named "Car part", shouldn't we also CamelCase All Words? Or too much magic?

tisto:
@davilima6 I think we should at least make sure the type name starts with an upper case letter "car part" -> "Carpart". I guess we could enforce camel case but we would need proper testing for that because I could imagine a few edge cases.

faustram:
I'm sorry. 
I had a similar error like @pigeonflight. 
I named my add-on "PFinance.SingleMovement" and in interfaces.py, row14 I have:
"class IPFinance.SingleMovement(Interface):"
So "bin/instance fg" fail.

pbauer:
@faustram the Content type name should not contain space, dot, minus or special characters. We could add a check regex to check for valid names. See https://github.com/plone/bobtemplates.plone/pull/112

PR #82, opened 2015-08-08
fix issue #81 space in names breaks generated packages

pigeonflight:
This fix ensures that if someone gives a dexterity content a name with a space in it, the internal (hooks.py) code removes spaces from the resulting name.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changelog
 0.12 (unreleased)
 -----------------
 
+- Added support for Dexterity content types with spaces in their names
+  [pigeonflight]
+
 - Fix #84 Make travis cache the egg directory of the generated package.
   [jensens]
 
@@ -12,6 +15,7 @@ Changelog
 
 - Remove unittest2 dependency.
   [gforcada]
+
 
 0.11 (2015-07-24)
 -----------------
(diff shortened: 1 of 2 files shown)


tisto:
@pigeonflight thanks for the fix! Looks good. Though, it seems the tests fail and it would be great if we could add a test for this as well. Would you mind having a look?

tisto:
@pigeonflight it seems there is still one flake8 violation: 

/home/travis/build/plone/bobtemplates.plone/bobtemplates/hooks.py:189:17: F821 undefined name 'dexterity_type_name'

pbauer:
@pigeonflight You need to use dexterity_type_name_lower in several other places: interfaces.py, fti-xml /filename and content), the types.xml and maybe others. But you'' end up with a type class `class demotype()` instead of `class DemoType` so you might want to not user lower()...

pbauer:
:-1:  imho type-names with spaces are no good idea. The News Item created many annoying issues (the latest being https://github.com/plone/plone.app.contenttypes/commit/3f209d737b10504bf6fac794036a267484015a3e)

tisto:
No activity here since quite a while. Will close the issue now. Feel free to re-open if you disagree...

PR #86, opened 2015-08-10
fix #84 Make travis cache the egg directory of the generated package.

### LINKED ISSUE #84: Make travis cache the egg directory of the generated package.
This could significantly speed up the travis build time. 8 minutes per build is too much.

jensens:
see #84

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,12 @@ cache:
   pip: true
   directories:
     - $HOME/buildout-cache
-    - $HOME/test.plone_addon/eggs
 before_install:
   - mkdir -p $HOME/buildout-cache/{eggs,downloads}
+  - mkdir $HOME/.buildout
+  - echo "[buildout]" > $HOME/.buildout/default.cfg
+  - echo "download-cache = $HOME/buildout-cache/downloads" >> $HOME/.buildout/default.cfg
+  - echo "eggs-directory = $HOME/buildout-cache/eggs" >> $HOME/.buildout/default.cfg
   - virtualenv .
   - bin/pip install --upgrade pip setuptools zc.buildout
 install:
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changelog
 0.12 (unreleased)
 -----------------
 
+- Fix #84 Make travis cache the egg directory of the generated package.
+  [jensens]
+
 - Update tests to use Plone 5.0b3.
   [jensens]
 
--- a/addon_test.template.sh
+++ b/addon_test.template.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+# just a  template, used by buildout to generate a test script to run tests
+# of the addon
+set -e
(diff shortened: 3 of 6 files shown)


tisto:
@jensens not exactly what I meant but nice anyways. :)

I was thinking about a way to speed up the travis job for bobtemplates.plone itself. We currently cache the eggs of the main buildout but the tests create a new package buildout that takes time to run and that does not use any egg caching so far.

jensens:
ops :)

jensens:
but wait, since both are running on the same travis and both are using the absolute path in travis.cfg this will fix it too, right?

jensens:
it does not fix it, because it runs buildout.cfg - not travis.cfg - in the tests

jensens:
sorry, i'am trying to find the place where (in code) do the tests create a new package buildout?

jensens:
@tisto @ pbauer this one is ready for merge, review appreciated, build/test time on travis reduced from 8 to 9 min per variant down to ~1:30 with warmed up cache.

tisto:
@jensens that's great! The reason why I haven't merged the pr yet is that I'm a bit concerned about the complexity of our overall build process. I'm wondering if would be possible to use a ".buildout" egg cache in the travis home root instead of injecting the cache into the buildout configuration.

jensens:
hmm, do you mean to add a global ``$home/.buildout/default.cfg`` like this?

    [buildout]
    download-cache = /home/travis/buildout-cache/downloads
    eggs-directory = /home/travis/buildout-cache/eggs

I think this could work and is much simpler. I give it a try.

tisto:
@jensens yes. this is what I had in mind.

jensens:
indeed, much simpler this way. I just did not thought about the possibility of global configuration.

jensens:
@tisto ok, all green, I squashed it to one commit - if tests are still green (no reason why not) its imo really ready for merge - and imo this way a very elegant solution.

tisto:
AWESOME! Feel free to merge when the build is green.

mauritsvanrees:
The Travis jobs for bobtemplates.plone are fast, so this apparently works, but I wonder how. I see one problem:

- We now have a `$HOME/.buildout/default.cfg` that says `eggs-directory = $HOME/buildout-cache/eggs`
- The travis.cfg of the generated package ultimately extends https://raw.githubusercontent.com/collective/buildout.plonetest/master/buildout-cache.cfg which says `eggs-directory = ${buildout:directory}/buildout-cache/eggs`

So it looks to me like the generated package listens to `buildout-cache.cfg` and will start creating eggs in `/home/travis/build/plone/bobtemplates.plone/test.plone_addon/buildout-cache/eggs`
It is faster than it is supposed to. Am I missing something?

mauritsvanrees:
Note: the label cleanup I just did on this issue has nothing to do with my comment, but pr review has been done and pr ready seems unneeded since this was merged already. Just brushing some corners. :-)

PR #88, opened 2015-08-11
Install before use

gforcada:
One needs to install bobtemplates.plone before being able to create a package.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -129,9 +129,11 @@ Installation in a virtualenv
 
 You can also install ``bobtemplates.plone`` in a virtualenv.::
 
-    $ pip install mr.bob
-
     $ pip install bobtemplates.plone
+
+With ``pip 6.0`` or newer ``mr.bob`` will automatically be installed as a dependency. If you still use a older version of pip you need install ``mr.bob`` before ``bobtemplates.plone``.::
+
+    $ pip install mr.bob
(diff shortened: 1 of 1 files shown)


tisto:
@gforcada Maybe we should just reorder the sections to "installation", "usage" and "options."

gforcada:
Oh I see, I never scrolled that much :sorry: then nevermind

tisto:
@gforcada No, no. I think you are right. The way the readme is written is confusing. We should fix that.

pbauer:
Also you cannot install bobtemplates.plone without first installing mr.bob. Installing mr.bob as a dependency does not work. See https://github.com/domenkozar/mr.bob/issues/70

tisto:
@pbauer that was not the point (I think). The readme uses mr.bob/bobtemplates.plone before installing it. This is confusing for the users because they have to read the entire readme to understand how to use bobtemplates.plone.

gforcada:
@pbauer it did work for me:

```
mkdir test
cd test
virtualenv-2.7 .
source bin/activate
pip install bobtemplates.plone
mrbob -O collective.addon bobtemplates:plone_addon
```

So maybe it's already resolved?

do3cc:
This documentation can be updated, because it works now without installing first

PR #97, opened 2015-09-10
Use Plone minor version in classifiers, and various other small fixes

mauritsvanrees:
Too small for separate pull requests, and they all look safe to me, but this is my first contribution to bobtemplates.plone, so someone might as well have a look at the bunch.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,10 @@ Changelog
 0.12 (unreleased)
 -----------------
 
+- Use Plone minor version for ``setup.py`` classifier.  So 4.3 instead
+  of 4.3.6.
+  [maurits]
+
 - Enabled robot part in generated package.
   [maurits]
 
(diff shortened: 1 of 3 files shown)


mauritsvanrees:
Travis does not like it, and it has a point. Apparently there is a bin/nosetests that I should have run. Don't tell me I needed to read some documentation... ;-)

nosetests_answers.ini
```diff
@@ -15,4 +15,5 @@ author.irc = irc.freenode.org#plone
 
 plone.version = 4.3.6
 plone.is_plone5 = False
+plone.minor_version = 4.3
```
tisto: I think we can extract the minor version from plone.version without adding an additional question.

tisto:
I guess we should create a wrapper "test" script that runs all acceptance tests and nosetests.

nosetests_answers.ini
mauritsvanrees: We *are* extracting it from plone.version. But I guess when you use a .ini file, the code that extracts it (`post_plone_version` in `hooks.py`) is not getting called.

mauritsvanrees:
I need to revisit this for a minor update once pull request #101 is merged.

mauritsvanrees:
I rebased. Tests now pass. Coverage is not working yet because that is in a different pull request.
Ready for review.

PR #99, opened 2015-09-16
Cleanup generated MANIFEST.in, add setup.cfg.

mauritsvanrees:
There is no need to include *cfg files, nothing is done with it.

The MANIFEST.in file is included automatically, so no need to spell it out.

We must globally exclude *.pyc.

In short, the check-manifest command should not complain.  This is a
stand-alone command and an optional helper package for zest.releaser,
called during prerelease.

Also, this fixes an error on Windows, noticed by check-manifest:
ERROR: Trailing slashes are not allowed in MANIFEST.in on Windows: docs/

### DIFF
--- a/bobtemplates/plone_addon/MANIFEST.in.bob
+++ b/bobtemplates/plone_addon/MANIFEST.in.bob
@@ -1,3 +1,4 @@
 graft src/{{{ package.namespace }}}
-graft docs/
-include *.cfg *.rst *.in
+graft docs
+include *.rst
+global-exclude *.pyc
--- a/bobtemplates/plone_addon/setup.cfg.bob
+++ b/bobtemplates/plone_addon/setup.cfg.bob
@@ -1,0 +1,7 @@
+[check-manifest]
+ignore =
+    *.cfg
+    .coveragerc
+    .editorconfig
+    .gitattributes
+    bootstrap-buildout.py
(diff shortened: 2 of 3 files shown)


mauritsvanrees:
No idea yet why travis (and local tests) give: `KeyError: 'profile-plone.app.event:default'`. Should have nothing to do with this pull request.

tisto:
It seems plone.app.event 2.x is fetched and this is breaking the build. Grrr...

mauritsvanrees:
I do at least need to fix some tests that *are* definitely related to this pull request.

I'll have a look at the plone.app.event thingie too.

tisto:
Maintaining a Plone 4.3.x buildout with p.a.event is really a horrible experience. It breaks with every new release...now we even have to pin plone.formwidget.recurrence to make it work.

https://github.com/plone/bobtemplates.plone/pull/100

@thet let's please fix that for the next Plone 4.x release.

tisto:
Pinning the versions in buildout won't work since we need to test against Plone 5.x and 4.x at the same time. I don't think it makes sense to fix this here. We have to clean up our p.a.event dependency hell.

mauritsvanrees:
I rebased. Tests now pass. Coverage is not working yet because that is in a different pull request.
Ready for review.

PR #100, opened 2015-09-16
Pin p.a.contenttypes, p.a.event, and p.f.recurrence.

tisto:


do3cc:
:-(

do3cc:
@esteele This is a bug in the versions.cfg on dis.plone.org is there no way to add these fixes there?

tisto:
@do3cc we should pin all those versions in the next 4.3.7 release. Though, bobtemplates.plone needs to support older 4.3.x versions anyways. Therefore I guess there is no way around pinning those packages in our buildout for a while.

PR #101, opened 2015-09-16
Generated buildout: no plone.app.event 2 on Plone 4.3.

mauritsvanrees:
Pin plone.app.event to 1.2.7 on Plone 4.3.

Should be done nicer, but let's first see if this fixes travis failures.

See problems at pull request #99.

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -89,9 +89,6 @@ class PloneTemplateTest(BaseTemplateTest):
                 self.project + '/src/collective/foo/profiles/default/browserlayer.xml',  # noqa
                 self.project + '/src/collective/foo/profiles/default/collectivefoo_default.txt',  # noqa
                 self.project + '/src/collective/foo/profiles/default/metadata.xml',  # noqa
-                self.project + '/src/collective/foo/profiles/uninstall',  # noqa
-                self.project + '/src/collective/foo/profiles/uninstall/browserlayer.xml',  # noqa
-                self.project + '/src/collective/foo/profiles/uninstall/collectivefoo_uninstall.txt',  # noqa
                 self.project + '/src/collective/foo/setuphandlers.py',
                 self.project + '/src/collective/foo/testing.py',
                 self.project + '/src/collective/foo/tests',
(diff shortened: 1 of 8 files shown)


tisto:
I really would like to avoid things like this. We are fixing the symptoms and not the source of the problem.

mauritsvanrees:
Agreed. This might be okay as a temporary fix, if it works.

tisto:
@mauritsvanrees fyi: 4.3.x works with those pins: https://github.com/plone/bobtemplates.plone/pull/100/files

mauritsvanrees:
In our `test_answers.ini` we have `plone.is_plone = False`. `mr.bob` interprets this as a string instead of a boolean, so it is interpreted as True... So this needs a fix in mr.bob.

Also, in Travis we are updating the `plone.version` in the test answers, but not the `plone.is_plone5` and `plone.minor_version` answers. Ideally, we would not even specify them, but `mr.bob` would calculate them for us.

mauritsvanrees:
The way we ran the tests, we were always generating Plone 5 templates... That's why it failed.

tests.py
```diff
@@ -89,9 +89,6 @@ def test_plone_addon_template(self):
                 self.project + '/src/collective/foo/profiles/default/browserlayer.xml',  # noqa
                 self.project + '/src/collective/foo/profiles/default/collectivefoo_default.txt',  # noqa
                 self.project + '/src/collective/foo/profiles/default/metadata.xml',  # noqa
-                self.project + '/src/collective/foo/profiles/uninstall',  # noqa
```
tisto: @mauritsvanrees I'm wondering why uninstall has been removed here...need to check...

tisto:
@mauritsvanrees awesome! Thanks for cleaning up! :)

tests.py
mauritsvanrees: `profiles/uninstall` is only created for Plone 5 packages. Or rather, when `plone.is_plone5` is False, we delete it afterwards. Until now our answer to that question in the tests was always wrongly interpreted to mean True, so it got generated.

Probably the uninstall profile should simply always be generated. IIRC, on Plone 4 we would need an extra uninstall method in `Extensions`, and on Plone 5 that is not needed.

Issue #93, opened 2015-08-13
Coverage/Coveralls does not grip.


jensens:
For some reason it does not show coverage results at  https://coveralls.io/github/plone/bobtemplates.plone

mauritsvanrees:
Looks like coveralls has never run for us.

In our `.travis.yml` we are not creating a coverage report. So we should call `bin/createcoverage`. But it we add this part, like we do for the addons we create, and we call it, it quits immediately because it expects there to be a `bin/test` that is a python script, and for us it is a shell script.

I briefly tried the next python code for bin/test, but then coverage tells us that we have 100% test coverage within the three lines of the`bin/test` script itself, which is not quite what we are after...:

```
import subprocess
import sys
sys.exit(subprocess.call('./bin/nosetests'))
```

If I hack the createcoverage script to run `bin/nosetests`, it goes better. It even says 100% coverage, but, well, we are only testing our one line `__init__.py`....:

![schermafbeelding 2015-09-16 om 20 04 56](https://cloud.githubusercontent.com/assets/210587/9913716/794c488c-5cae-11e5-84fb-33551872338d.png)

Our `hooks.py` file is not tested at all like this, because it is not called.  The original `bin/test` script does call it, but in this setup we would only call `bin/nosetests`.

Anyway, if we want coverage reports, the first step would be:
add an option to the createcoverage recipe so we can tell it to run bin/nosetests instead of bin/test.

Or does anyone know an alternative to createcoverage that we can use?

mauritsvanrees:
Fixed.

PR #102, opened 2015-09-16
Create coverage report in Travis.

mauritsvanrees:
This should solve issue #93, in a slightly different way than my last proposal there.

Generate nosetests script also under the 'test' name.  Then
bin/createcoverage works.

In Travis call the new correct script names:

- The new bin/test is the old bin/nosetests.
- The new bin/test-plone_addon is the old bin/test.

Created bin/alltests script.  This runs bin/test, and the two
test-plone_addon scripts.  We could call that in Travis too.

Added initial simple test to show coverage in hooks.py.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,10 +39,11 @@ before_script:
 script:
   - bin/check-readme
   - bin/code-analysis
-  - bin/nosetests
   - bin/test
+  - bin/test-plone_addon
 
 after_success:
+  - bin/createcoverage
   - pip install -q coveralls
   - coveralls
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ install:
   - pip install docutils
   - sed -ie "s/^plone.version.*/plone.version = $PLONE_VERSION/g" test_answers.ini
   - sed -ie "s/^package.type.*/package.type = $PACKAGE_TYPE/g" test_answers.ini
-  - if [ "$PACKAGE_TYPE" == "Dexterity" ]; then echo "package.dexterity_type_name = MyDextrityTestType" >> test_answers.ini; fi
+  - if [ "$PACKAGE_TYPE" == "Dexterity" ]; then echo "package.dexterity_type_name = MyDexterityTestType" >> test_answers.ini; fi
   - bin/buildout -c travis.cfg
 
 before_script:
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[report]
+include =
+  bobtemplates/*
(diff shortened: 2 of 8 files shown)


mauritsvanrees:
Travis is expected to fail, see a different pull request for the plone5 /plone.app.event stuff.
But after it is done, coveralls should hopefully get some data.

mauritsvanrees:
Yes!  The Plone 5 builds work and they pass info to coveralls.
The Plone 4 builds fail, so they do not contact coveralls, but we know that they fail.
Ready for review.

.travis.yml
```diff
@@ -39,8 +39,9 @@ before_script:
 script:
   - bin/check-readme
   - bin/code-analysis
-  - bin/nosetests
   - bin/test
+  - bin/createcoverage
```
tisto: I think createcoverage should run in the after_success sestion. This is not a test.

all_tests.template.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
tisto: Personally I prefer to keep such small scripts inline in the buildout. Inline templates are much easier to understand. Otherwise people have to look into many different files to understand what goes on.

full_test.py.template
```diff
@@ -0,0 +1,3 @@
+import subprocess
```
tisto: Can you elaborate where this file is actually used?

tisto:
@mauritsvanrees would you prefer to fix the Plone 4 build in this branch or shall we merge first?

.travis.yml
mauritsvanrees: Problem is that it then runs after `bin/test-plone_addon`, which leaves behind a `test.plone_addon` directory that the nosetests then also get run on, causing ImportErrors. Ah, but of course this directory only gets left behind when the command fails, so then there is no need to call createcoverage.  I will move it.

In that case, we should move createcoverage in the generated package too. I will move that too.

all_tests.template.sh
mauritsvanrees: Ah, true. Will do.

full_test.py.template
mauritsvanrees: Oops, this was a left-over from an earlier try. Removed now.

mauritsvanrees:
Plone 4 fixes have been merged to master.
I have fixed the issues you raised on this pull request, rebased on master, and did a force push, cleaning things up. Let's see what Travis and Coveralls think of it.

mauritsvanrees:
Two minor changes added. Let me stop now. :-)

tisto:
Thanks so much @mauritsvanrees! Great work! Will do a release soon.

PR #111, opened 2015-12-22
change browser layer name

thet:
Change the browser layer from the camel case'd package name to IBrowserLayer. This is easier to remember, more consistent and faster to write. The interface is prefixed by the package namespace anyways.

This is actual my personal habit - I switched from package-individual names to the name ``IThemeLayer`` because of the reasons mentioned above. "Theme" is maybe to opinionated, so I choose "Browser" in this PR, because that's what the interface is - a browser layer.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,9 @@ Changelog
 - Update ``.gitignores`` in repository to exclude ``lib64``, ``pip-selfcheck.json`` and all ``.*`` except necessary.
   Update ``.gitignore.bob`` in templates with these changes too.
   Add ``.gitattributes`` in repository for union-merge CHANGES.rst files.
+- Change the browser layer from the camel case'd package name to
+  ``IBrowserLayer``. This is easier to remember and faster to write. The 
+  interface is prefixed by the package namespace anyways.
   [thet]
 
 - Update robot test framework versions including Selenium to work with recent
--- a/bobtemplates/hooks.py
+++ b/bobtemplates/hooks.py
@@ -179,9 +179,10 @@ def prepare_render(configurator):
     camelcasename = dottedname.replace('.', ' ').title()\
         .replace(' ', '')\
         .replace('_', '')
-    browserlayer = "{0}Layer".format(camelcasename)
 
-    # package.browserlayer = 'CollectiveFooSomethingLayer'
+    browserlayer = "IBrowserLayer"
+
+    # package.browserlayer = 'IBrowserLayer'
     configurator.variables['package.browserlayer'] = browserlayer
 
     # package.longname = 'collectivefoosomething'
--- a/bobtemplates/hooks.py
+++ b/bobtemplates/hooks.py
@@ -176,14 +176,14 @@ def prepare_render(configurator):
         'package.dottedname'
     ].replace('.', '_').upper()
 
+    browserlayer = "BrowserLayer"
+
+    # package.browserlayer = 'BrowserLayer'
+    configurator.variables['package.browserlayer'] = browserlayer
+
     camelcasename = dottedname.replace('.', ' ').title()\
         .replace(' ', '')\
         .replace('_', '')
-
-    browserlayer = "IBrowserLayer"
-
-    # package.browserlayer = 'IBrowserLayer'
-    configurator.variables['package.browserlayer'] = browserlayer
 
     # package.longname = 'collectivefoosomething'
     configurator.variables['package.longname'] = camelcasename.lower()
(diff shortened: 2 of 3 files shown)


pbauer:
If we do this there is no need to generate it dynamically with ``I{{{ package.browserlayer }}}`` since it will alsways be ``IBrowserLayer``.

thet:
That's true. A reason to keep it in a variable would be, that it's convenient to have it in a variable, so that it can easily be changed... Not sure about that...
I can update the pull and change all references to the static IBrowserLayer.

thet:
hmhm... we can also rename the base testing layer class and instances and fixtures in testing.py to something more generic. except the layer instance names...
i'll give another try on that later...

tisto:
Variable names in general should have specific names in my opinion. This is not just a browser layer we are talking about but a package specific browser layer. If we remove that specific naming and replace it with a convention, it is essential that everybody knows about this convention and uses it (which is not the case afaik, and we are not going to change this in all packages). Therefore this change will lead to more confusion than it will help in my opinion. I could also imagine confusions and problems when the full import path is not available (debugging, error traces, inheriting, etc.).

jensens:
-1 I agree with @tisto

thet:
The motivation for this pull request was, that I generated a package with the namespace ``bda.aaf.site`` and the browser layer interface was named ``IBdaAafSiteLayer``.

Besides from looking ugly I don't want to have to remember that complicated name.
Normally you only refer to browser layers in other packages in zcml, and there you have give the full dotted path to them always anyways.

Consequently, if we want to have unambiguous names, we'd have to name our behavior interfaces, content type classes and the like also like that, e.g. ``IBdaAafSiteStunningBehavior``. But we don't want that.

If ``IBrowserLayer`` is to confusing, we could name it ``IPackageLayer`` also.

Currently the score is at 2:2, if I count @pbauer comment as pro this pull request :)
I'd like to hear another voice, so that the score is unequivocal.

However, I think for own projects, I'd have to subclass bobtemplates.plone for our specific conventions anyways. Not only for the naming, but also own CI setups and so forth.

gforcada:
@thet change the default value for a corner case doesn't look right to me.

I agree with you that ``IBdaAafSiteLayer`` looks ugly, but after generating the skeleton with bobtemplates.plone nothing prevents you from doing some renaming here and there.

As long as bobtemplates.plone only creates the initial skeleton and is not able to add things later on it's totally safe to just do some manual renaming if things look ugly.

As long as bobtemplates.plone is fully consistent the renaming should be safe as well.

jensens:
@gforcada I dont think this is a real corner case. Lots of customer specific packages are created in a third level namespace trying to avoid long names at first and second level.

And I agree that generic names are not good either. I really like to have a self-speaking name over some generic naming, also for readability reasons.

What I do in practice is to use a shortened name of the package, i.e. use "pact" for plone.app.contenttypes or customer specific "kup.stadtentwicklung.innsbruck" turns into "Stadtibk" which I use as browser layer as "IStadtibkLayer" and tests are named "STADTIBK_FUNCTIONAL_TESTING". To be used only for cases where already a namespace is in context, otherwise namespace clashes are possible.

For @thet his case a shortname like "aafsite" would turn into "IAafsiteLayer" or "AAFSITE_FUNCTIONAL_TESTING".

This would also avoid overlong names which then ends up breaking code-analysis and getting a ``# noqa`` at the end which is not good either.

I this approach would get some +1 I'd propose to ask for an optional short-name in bobtemplates.plone, where the default just points to the camel-cased namespace.

Opinions?

tisto:
@jensens no more options please. It is already quite hard to tests things properly. More options mean more (untested) code complexity.

jensens:
@tisto thats true. so lets keep it as is, manual change afterwards is probably easiest? (``rpl`` is my friend here)

tisto:
Will close the issue for now. Feel free to re-open if someone disagrees...

mauritsvanrees:
@jensens Thanks for pointing at [rpl](http://www.laffeycomputer.com/rpl.html) for replacing strings in files. I used my own hacked together shell script to far. :-)

iham:
even if its closed:

"some renaming" reminds me, what mrbob currently is to me:
a skeleton which result i have to question as its not what i am going to use.
coming from zopeskel, paster, templer - even knowing that mrbob is not like that - i still want a similar result: a working skeleton i dont have to refactor.

when creating a theme or dexterity type i don't want to rename stuff. especially if its as basic as the browserlayer. renaming leads always to errors. so a good name would be all it needs.

why is IBrowserLayer unreadable? as it is exactly that for a theme.
i can only asume using it as a import. so the import is speaking.
from my.theme.interfaces import IBrowserLayer
from my.dx.type.interfaces import IBrowserLayer

if using multiple imports in one module i used "as" and lets say "IMyThemeBrowserLayer"

why do dexterity types even need a browserlayer?
if they do: using "ibrowserlayer" for both would be good as it is correct for both.
only naming it themelayer for a theme is kind of ok-ish too.
but i prefer to not separate the names if it is the same habit.

PR #112, opened 2015-12-30
Validate type name input (fixes #81)

### LINKED ISSUE #81: Does not deal well when creating Dexterity content types with two words

pbauer:


bobtemplates/hooks.py
```diff
@@ -141,6 +143,14 @@ def pre_dexterity_type_name(configurator, question):
         raise SkipQuestion
 
 
+def post_dexterity_type_name(configurator, question, answer):
```
tisto: @pbauer wouldn't it make sense to write a few unit tests to show how this is supposed to work and to make sure the function works as expected?

tisto:
@pbauer what's the status of this pr? Would you mind rebasing and adding a few tests?

pbauer:
@tisto done

tests.py
```diff
@@ -182,3 +183,27 @@ class HooksTest(unittest.TestCase):
     def test_to_boolean(self):
         # Initial simple test to show coverage in hooks.py.
         self.assertEqual(hooks.to_boolean(None, None, 'y'), True)
+
+    def test_post_dexterity_type_name(self):
+        """Test validation of extered dx type names
```
gforcada: typo *extered* :-)

PR #115, opened 2016-01-13
Update robot test framework versions

thet:
Update robot test framework versions including Selenium to work with recent firefox releases.

Otherwise it fails with:
```
Failure in test Scenario As a member I want to be able to log into the website (test_example.robot)
Traceback (most recent call last):
  File "/home/thet-data/dotfiles-thet/home/.buildout/eggs/unittest2-0.5.1-py2.7.egg/unittest2/case.py", line 340, in run
    testMethod()
  File "/home/thet-data/dotfiles-thet/home/.buildout/eggs/robotsuite-1.6.1-py2.7.egg/robotsuite/__init__.py", line 455, in runTest
    assert last_status == 'PASS', last_message
AssertionError: Setup failed:
WebDriverException: Message: Can't load the profile. Profile Dir: %s If you specified a log_file in the FirefoxBinary constructor, check it for details.
```


thet:
Probably only Selenium is the problem...

jensens:
+1 - had the same problem here

jensens:
btw it is really annoying selenium has so often problems after ff update 
</rant>

PR #116, opened 2016-02-05
gitignores and gitattributes

thet:
Update ``.gitignores`` in repository to exclude ``lib64``,
``pip-selfcheck.json`` and all ``.*`` except necessary.
Update ``.gitignore.bob`` in templates with these changes too.
Add ``.gitattributes`` in repository for union-merge CHANGES.rst files.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+CHANGES.rst merge=union
--- a/.gitignore
+++ b/.gitignore
@@ -1,13 +1,9 @@
-!src/bobtemplates
+# any of...
+.*
 *.egg-info
 *.mo
 *.pyc
-.coverage
-.installed.cfg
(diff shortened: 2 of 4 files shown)


fgrcon:
Is it really a good idea to ignore **src/***  in addon-templates  -  or. maybe I missed something?

do3cc:
@fgrcon look at the last line of gitignore, where the specific source package is readded. This can be useful if you add other source packages via mr.developer, which are often also checked out into the src directory. This would avoid that these checkouts mark the repo as dirty.

do3cc:
@thet otoh, it would be better to only include __init__.py from the namespace dir and then only the package itself. else a package collective.foo with a mr.developer checkout of collective.solr would mark the git repo as dirty too.

PR #119, opened 2016-02-10
Add "(uninstall)" to the uninstall profile title.

thet:
Otherwise it cannot be distinguished from the install profile in portal_setup.

By the way, why is the uninstall profile only created for Plone 5 and not Plone 4 also?


jensens:
+1

PR #122, opened 2016-02-18
Complete the theme.xml GenericSetup registry file with an a xmlheader…

thet:
… and a themes tag.

The ``theme.xml`` fell somehow out of the convention we follow for other GS xml files. No xml header tag and how to add more than one theme?

We should follow best practices also here, if this is best practice.

However, disabling a theme with ``<enabled>false</enabled>`` (as in the uninstall profile) doesn't realy work - we'd had to enable another one instead. Enabling barceloneta seems also not to work. However, these are plone.app.theming issues.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changelog
 1.0.2 (unreleased)
 ------------------
 
+- Complete the ``theme.xml`` GenericSetup registry file with an a xmlheader and a ``themes`` tag.
+  [thet]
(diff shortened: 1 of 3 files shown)


thet:
DO NOT MERGE!
does not work.
have to investigate plone.app.theming GS handler.

thet:
Does not work like that. plone.app.theming's theme.xml is only to activate a theme, not to configure multiple ones.

PR #123, opened 2016-02-18
Add encoding to all GS xml declarations

thet:
Declare the xml encoding for all GenericSetup profile files.
Otherwise the parser has to autodetect it.
Also add an xml version and encoding declaration to ``theme.xml``.

/cc @jensens thinking this further, shouldn't ZCML files also get this xml declaration heading ``<?xml version="1.0" encoding="UTF-8"?>``?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,11 @@ Changelog
 1.0.2 (unreleased)
 ------------------
 
+- Declare the xml encoding for all GenericSetup profile files.
+  Otherwise the parser has to autodetect it.
+  Also add an xml version and encoding declaration to ``theme.xml``.
+  [thet]
+
 - Add "(uninstall)" to the uninstall profile title.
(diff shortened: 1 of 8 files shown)


jensens:
valid XML should always have the header, but not all parsers are working with it.

gforcada:
@thet note that the tests that fail are only the 4.3.x ones with the following traceback:

```
The constraint, 4.3.7, is not consistent with the requirement, 'plone.app.locales>=4.3.9'.

While:

  Installing instance.

Error: Bad constraint 4.3.7 plone.app.locales>=4.3.9
```

thet:
Testfailures are unrelated, looks fine to me! ;)

tisto:
@thet merging into a broken build is never ok. No matter if the changes in the pr are related or not. Please fix the master first if you want to merge stuff or wait until somebody else does it. CI rules apply no matter if we are on Jenkins or on Travis.

PR #124, opened 2016-02-24
Remove the ``xmlns:five`` namespace, as it is not used at all.

thet:


gforcada:
We need a plone.recipe.codeanalysis check for this kind of things :-)

do3cc:
I don't know but maybe we should keep the common namespaces for convenience. They do not add a penalty to load times.

thet:
Many years back ago, when I was more unexperienced than now, I was wondering what such magic code lines had to mean. I could have asked myself "``xmlns:five="http://namespaces.zope.org/five"`` for sure pulls in something very important for my application to work, only I don't know what...." without knowing that this is just a unused xml namespace.
I'd leave out what is not necessary, just to avoid such confusion....

do3cc:
I am unsure, just because I can never remember the urls, I guess everybody greps for it in existing source code, that is not something I'd like to teach as best practice in a developer training.
OTOH we can probably get rid of that namespace anyway, now that we control the full stack :-)

PR #126, opened 2016-02-29
Fix build failure on 4.3 with testing extra

thet:
Fixes issue discussed in https://github.com/plone/bobtemplates.plone/pull/123
Rebases on https://github.com/plone/bobtemplates.plone/pull/125 which is a prerequisite.
@tisto plz review

/cc @gforcada


do3cc:
Should we not make a new release of plone.app.contenttypes without that breaking change?

gforcada:
Looks good, the p.a.contenttypes doesn't feel right somehow, but probably once we get a 4.3.8(?) that will be fixed?

thet:
These packages have a dependency on plone.app.locales>=4.3.9:

plone.formwidget.datetime
plone.app.lockingbehavior
plone.app.referenceablebehavior

There is a missing version definition for plone.formwidget.datetime in http://dist.plone.org/release/4.3.7/versions.cfg , while plone.app.contenttypes and plone.app.event have one. So this might be the cause.

I agree, this needs fixing in the next Plone 4.3 release by adding a plone.formwidget.datetime dependency in buildout.coredev version.cfg.

Any takers? I'm offline for at least the next 12 hrs.

mauritsvanrees:
plone.formwidget.datetime only has Plone 5 in its setup.py classifiers, but I was the one who added this last year as this package was only in the sources.cfg in Plone 5. But it should work fine in Plone 4.3. Seems fine to add the pin there. I'll do that.
So that does not help currently, but it will help in future similar cases.

thet:
@mauritsvanrees any plans when to release 4.3.8?
Or should we merge this PR?

mauritsvanrees:
Plone 4.3.8 is swiftly coming, but meanwhile this seems fine. It fixes current Travis failures. I'll merge.

PR #127, opened 2016-03-02
Note about disabled z3c.autoinclude in test layer setup.

thet:
... Took me half a day to find this out. Hope that helps others.

### DIFF
--- a/bobtemplates/plone_addon/src/+package.namespace+/+package.name+/testing.py.bob
+++ b/bobtemplates/plone_addon/src/+package.namespace+/+package.name+/testing.py.bob
@@ -25,8 +25,8 @@ class {{{ package.browserlayer }}}(PloneSandboxLayer):
 {{% endif %}}
 
     def setUpZope(self, app, configurationContext):
-        # Load any other ZCML from your package' dependencies manually here.
-        # The z3c.autoinclude feature is disabled for the Plone fixture base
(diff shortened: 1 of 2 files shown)


bobtemplates/plone_addon/src/+package.namespace+/+package.name+/testing.py.bob
```diff
@@ -25,6 +25,9 @@ class {{{ package.browserlayer }}}(PloneSandboxLayer):
 {{% endif %}}
 
     def setUpZope(self, app, configurationContext):
+        # Load any other ZCML from your package' dependencies manually here.
```
do3cc: Load any other ZCML that is required for your tests...
Z3c.autoinclude is not disabled so that you have to add EVERYTHING manually

thet: The Plone fixture disables the autoinclude feature here:
https://github.com/plone/plone.app.testing/blob/master/plone/app/testing/layers.py#L134
but loads other ZCML in the following line...

I'll change to your suggested wording.

do3cc:
Just a quick pointer if you want to know more:
https://community.plone.org/t/what-is-best-practice-for-meta-zcml-files-and-test-layers-in-plone-app-testing/442

PR #128, opened 2016-03-24
Replaced import steps by post_handlers.

mauritsvanrees:
Needs GenericSetup 1.8.2 or higher (Plone 4.3.8 or 5.0.3).

In `.travis.yml` I have set the Plone version to 5.0.3-pending, otherwise we get a version conflict for GenericSetup.
I suggest we hold off a merge until 5.0.3 is officially released and we can properly set the version in `.travis.yml`.


mauritsvanrees:
I have updated the pull request to use Plone 5.0.3, since that was released yesterday on dist.plone.org.

tisto:
@mauritsvanrees if we merge this and the user chooses anything else than Plone 4.3.8 and 5.0.3 he/she will end up with a broken package with version conflicts, right?

jensens:
So a pinned GenericSetup package (and Quickinstaller?) is needed? 
My opinions on this subject differ, because 
a) I.Want.This.
b) I don't like that version pins are needed for a generated addon.

mauritsvanrees:
No new QuickInstaller is needed.

When you choose 4.3.7 and run the generated buildout, you indeed get an error:

```
The constraint, 1.7.7, is not consistent with the requirement, 'Products.GenericSetup>=1.8.2'.
While:
  Installing instance.
Error: Bad constraint 1.7.7 Products.GenericSetup>=1.8.2
```

Mostly I would find it strange if someone chooses a not current Plone version here, though it can happen.

We could work around this by adding a pin to `buildout.cfg.bob`. 1.8.2 is quite safe for any Plone 4.3.x site, and I think lower too.

We could add a hook that only adds the pin when needed.

Or would this be reason to reject this pull request because an integrator using the generated and released package would run into the same problem?

I would say this is a new best practice which should be added at some point. Question is when.
Options are also to document this, or make 4.3.8 and 5.0.3 the minimum accepted versions in the questions.

gforcada:
IMHO we should add it, if bobtemplates.plone is supposed to be the best practices skeleton generator, covering always latest releases and showcasing new approaches would be great.

do3cc:
zope.testing 5.0.0 ftw!1

do3cc:
Sorry, nearly nobody has the necessary context to understand my comment :-(
zope.testing 5.0.0 can seriously be used for bobtemplates packages. It contains no backward incompatible changes for fresh packages. It only breaks a full plone_coredev testrun, because it catches common errors users make. Because it caches these errors, it makes sense to use the newer version.

jensens:
I'am fine with directly supporting only the latest version and having to update pins for older versions.

do3cc:
As a maintainer of a few Plone Sites that do not yet run on the latest patch versions, I see it a tiny bit worrysome that new packages created with bobtemplates plone would not be compatible with my site but would not fail to install.
I also see the Products.GenericSetup does not declare a dependency on these newer versions.
Of what nature are these dependencies of GS 1.8.2 to Plone 5.0.3 or 4.3.8?

mauritsvanrees:
GenericSetup is used outside of Plone, so it does not declare a dependency on a Plone version.

The new Plone versions use the post_handlers that I introduced in GS 1.8.2.

Using 1.8.2 should be fine on earlier Plone versions.
GS 1.8.0 does introduce a slightly different way of handling dependencies. If you are hit by this difference then you have a corner case in an addon that arguably needs a fix in this addon.

BTW, I am biased: I have built most of the GS improvements the last two years or so, so I want to use the new stuff. :-)

do3cc:
If it works with older plones, I am fine, but your changelog entry strongly suggests otherwise.

do3cc:
My last comment was unclear. If it works with any plone 4.3.x, I am +1 with this if the changelog looses the requires 4.3.8 message.
If it only works with 4.3.8 and 5.0.3 I'd be +1 if the setup.py has these requirements too, so that people get notified about potential problems when running buildout and not, maybe the uninstall step of something.

mauritsvanrees:
Ah, good point, I could have said this clearer. I have rebased the branch and changed the wording.

PR #129, opened 2016-03-30
Removed `.*` from the `.gitignore` file.

mauritsvanrees:
This would ignore the `.gitkeep` files, which would mean some directories are not added when you do `git add` after generating a new project.

I had to fix this afterwards in a package I generated, where Zope failed to start up when a colleague tried it, because the `static` and `overrides` directories had not been added:
https://github.com/zestsoftware/collective.calltoaction/commit/1d1353a619354e6729e5ba30303146a427085481

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,3 @@
-# any of...
-.*
 *.egg-info
 *.mo
 *.pyc
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,10 @@ Changelog
 1.0.2 (unreleased)
 ------------------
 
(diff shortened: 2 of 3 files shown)


mauritsvanrees:
Small enough to merge myself.

mauritsvanrees:
I noticed that this means that `.coverage` is no longer ignored by git.  I added this back directly on master, in 98ce8a7deb655e2a33064023faba7e828a37a47b.

I see there are excludes in `.gitignore`, which might have been another way to solve this. I added `!.gitkeep` for good measure in that commit. So we could ignore `.*` again if we want to...

PR #132, opened 2016-04-01
plone testing 5.0.0 dependency

do3cc:
plone.testing 5.0.0 is backward compatible and helps users to avoid some common errors when writing tests.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -65,6 +65,10 @@ Changelog
 - Enabled robot part in generated package.
   [maurits]
 
+- Add depedency on plone.testing 5.0.0. Despite the major version number,
+  this change does not contain breaking changes.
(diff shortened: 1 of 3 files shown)


do3cc:
This PR is somehow bad, need to rebase properly

gforcada:
@do3cc as per your last comment (indeed the changelog entry needs to be moved up to the the unreleased section) I updated the labels, so @jensens doesn't merge it :-)

do3cc:
eek, it wasn't bad any more, at least so I thought

tisto:
@do3cc plone.testing itself does not have any requirements itself that could break buildout, right?

do3cc:
That is right. No new requirements. Only some tests may fail which didn't fail before. These tests were bad before.

PR #155, opened 2016-07-02
Explicitly load ZCML for locale overloading

thet:
Add the generated package explicitly to the buildouts instance zcml section to make sure it's loaded first.
That ensures that core locales can be overloaded by the package.

/cc @mauritsvanrees

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,10 @@ Changelog
 1.0.4 (unreleased)
 ------------------
 
+- Add the generated package explicitly to the buildouts instance zcml section to make sure it's loaded first.
+  That ensures, that core locales can be overloaded by the package.
+  [thet]
+
 - Pin versions of coverage/createcoverage
   [staeff]
 
--- a/bobtemplates/plone_addon/buildout.cfg.bob
+++ b/bobtemplates/plone_addon/buildout.cfg.bob
@@ -21,6 +21,10 @@ eggs =
     Pillow
     {{{ package.dottedname }}} [test]
 
+# To be able to overload core locales, our package zcml needs to be loaded first.
+zcml =
+    {{{ package.dottedname }}}
+
 
 [code-analysis]
 recipe = plone.recipe.codeanalysis[recommended]


tisto:
@thet the zcml slug is certainly useful when overriding locales. Though, I'm not sure if that makes a lot of sense for a single package buildout (and this is what bobtemplates.plone is right now). If we create a buildout template that contains a policy product, then it would make sense to add the zcml slug for that product. Though, I think this is currently beyond the scope of what mr bob can do. 

Maybe I'm missing something here but it seems to me we are just adding clutter here that will confuse people.

thet:
The buildout template in bobtemplates.plone is probably not the one which is used in real projects. But this PRs purpose is more about to document the necessity to add the package explicitly to the buildouts zcml section, if core locales should be overwritten.
It's also documented here, though: http://docs.plone.org/develop/plone/i18n/internationalisation.html#overriding-translations

davilima6:
I've been using the buildout.cfg that comes with the Dexterity template as a development buildout, which may serve as base when deployment is closer and a more comprehensive CFG is needed. So if I'm not missing anything I think this PR does make sense.

mauritsvanrees:
The extra zcml is fine for an individual project for one person or company, though you can also create a `locales` folder outside of the `src` and load it in the instance with `locales = ${buildout:directory}/locales`.

But I don't like this for an add-on meant to be used in multiple buildouts, probably by persons other than you. People will just add it to the eggs and and forget the zcml and then wonder why the changed translations don't show up.
Also, IMHO an add-on should not override core translations. In an individual project specific for a customer: sure. But not in a template that is meant to be used as basis for all add-ons.
So -1 from me.

Nothing really breaks when the zcml is explicitly loaded or not, but that can be used as argument both for and against this pull request...

thet:
@mauritsvanrees that does make sense...
Do we have a buildout-only template where this PR could make sense?
Otherwise, someone may please close this PR.

I just recognized that for me, the best place is the BDA organizations buildout base files:
https://github.com/bluedynamics/buildout-base/blob/master/etc/plone.cfg

PR #162, opened 2016-07-29
Get rid of travis.cfg configuration as its use is no longer considered best practice

hvelarde:
Refs. https://community.plone.org/t/clean-ish-way-to-get-travis-to-build-all-plone-versions/2539/8?u=hvelarde


hvelarde:
there are still some issues with code analysis, but it's late over here and I need some rest; do you mind to finish it?

.travis.yml
```diff
@@ -2,19 +2,16 @@ language: python
 python: 2.7
 sudo: false
 cache:
-  pip: true
   directories:
-    - $HOME/buildout-cache
+  - eggs
+  - test.plone_addon/eggs
```
hvelarde: @jensens there seem to be issues with the caching of this directory and I think they are caused because the name is the same in both cases.

.travis.yml
```diff
@@ -2,19 +2,15 @@ language: python
 python: 2.7
 sudo: false
 cache:
-  pip: true
   directories:
-    - $HOME/buildout-cache
+  - test.plone_addon/eggs
```
hvelarde: @jensens this seems not to be working in the context of the add-on test; we may probably want to add again the `default.cfg` file with an updated configuration or just get rid of the virtualenv inside the add-on test.

it's late now and I need to wake up "early"; check if you can finish this.

hvelarde:
@jensens there's still room for enhancement: the egg directory on the root level is not being cached as 2 directories can't share the same name and I decided to cache the one on the add-on.

maybe you can think on something around buildout default's configuration file; I'm done for today.

CHANGES.rst
```diff
@@ -4,6 +4,15 @@ Changelog
 1.0.5 (unreleased)
 ------------------
 
+- Keep versions of Robot Framework and friends centralized and updates.
```
hvelarde: updated...

hvelarde:
we need to debug further; for some reason the cache on the branch is empty: https://travis-ci.org/plone/bobtemplates.plone/caches

hvelarde:
I think the problem was this line: https://github.com/plone/bobtemplates.plone/blob/master/addon_test.template.sh#L24

travis.cfg
```diff
@@ -1,12 +0,0 @@
-[buildout]
-extends = buildout.cfg
-parts += createcoverage
-
-[code-analysis]
-recipe = plone.recipe.codeanalysis
-pre-commit-hook = False
-return-status-codes = True
```
tisto: @hvelarde if we remove travis.cfg p.r.codeanalysis won't be able to return proper status codes on travis.

tisto:
@hvelarde this pull request does a lot of things. Please consider splitting this into smaller pull requests to make it easier to discuss. I'm also not sure if we can get rid of travis.cfg because p.r.codeanalysis needs to return proper exit codes on travis to catch violations.

bobtemplates/plone_addon/src/+package.namespace+/+package.name+/tests/test_+package.dexterity_type_name_lower+.py.bob
```diff
@@ -1,15 +1,14 @@
...
-from plone import api
-
-from {{{ package.dottedname }}}.testing import {{{package.uppercasename}}}_INTEGRATION_TESTING  # noqa
 from {{{ package.dottedname }}}.interfaces import I{{{ package.dexterity_type_name }}}
+from {{{ package.dottedname }}}.testing import {{{package.uppercasename}}}_INTEGRATION_TESTING  # noqa
+from zope.component import createObject
+from zope.component import queryUtility
 
-import unittest2 as unittest
+import unittest
```
do3cc: \o/

travis.cfg
do3cc: He added it to general buildout.
Not sure why this isn't the default anyway.

hvelarde:
@tisto I separated the changes in many commits so you can review better; if you want, I can do different PR but I think is not necessary if you just merge it instead of squashing it.

I fixed code analysis as you can see from Travis; it was not running fine and the code had a couple of mistakes that where not detected before, like imports, quotes and plone.api usage; everything is clean now.

the thing that is still failing is the generation of cache in Travis (the core of this PR) and I don't understand why; the cache is empty as you can check in the link above.

I you have and clue, let me know. I may have some time later today to work a little bit on this again.

.travis.yml
jensens: possible, do we have some variable to append from the matrix?

.travis.yml
```diff
@@ -2,19 +2,16 @@ language: python
 python: 2.7
 sudo: false
 cache:
-  pip: true
```
jensens: why remove pip caching?

travis.cfg
jensens: +1 for code analysis in general buildout!

jensens:
Overall I like this Pull Request.
May you rebase/fixup the last two commits into the first?
Are any problems left? It's indeed more difficult to keep on track when multiple issues are addressed in one PR. But if the single commits are clean its ok imo.

travis.cfg
tisto: Code analysis is already in buildout.cfg, just with a different configuration. This discussion really gets out of hand, we are discussing 10 issues at the same time.

bobtemplates/plone_addon/buildout.cfg.bob
```diff
@@ -1,11 +1,14 @@
 [buildout]
-extends = http://dist.plone.org/release/{{{ plone.version }}}/versions.cfg
+extends =
+    http://dist.plone.org/release/{{{ plone.version }}}/versions.cfg
+    https://raw.githubusercontent.com/plone/plone.app.robotframework/master/versions.cfg
```
tisto: @hvelarde I don't think this is a good idea. This means buildouts generated by bobtemplates.plone can break at any time if something changes in p.a.robotframework repo. It also means that our travis build can fail at any time without us noticing.

tisto:
@hvelarde please close this pr and create a separate pr for every single issue. This makes it a lot easier to discuss and review the pull requests.

jensens:
Looking at the discussion I finally I agree with @tisto. Some of your changes could be merged immediately w/o discussion, others need some more work. Splitting it up makes this easier.

PR #163, opened 2016-08-01
Use selenium v2.53.6

hvelarde:


tisto:
@hvelarde as written before. I think this makes our build a lot more fragile. We won't be able to catch problems when versions change and at some point we will break all existing buildouts that depend on plone.app.robotframework. This is an anti-pattern for software development and should be avoided. I would suggest to close this pr.

hvelarde:
well, the version of Selenium you are using right now causes failures in current versions of Firefox; if you don't want to keep the whole thing updated I can change the PR to at least fix that.

tisto:
@hvelarde feel free to upgrade the selenium versions. Though, I would recommend to do a new pull request.

PR #164, opened 2016-08-01
Update .coveragerc

hvelarde:


.coveragerc
```diff
@@ -1,3 +1,7 @@
 [report]
-include =
```
tisto: @hvelarde could you elaborate why this didn't work? An omit statement is harder to maintain if people add stuff. An include is a better solution in my option (only if it works of course).

hvelarde: let me make a few tests; I think the right option is neither one, but [source](https://coverage.readthedocs.io/en/coverage-4.2/source.html).

hvelarde:
@tisto I tested `source` but it did not work so `include` seems to be fine; may I ask why you are omitting `test` and `upgrades`?

tisto:
@hvelarde the only thing this pull request now does is excluding the __init__.py on the first level of the package. Not sure why we should do that.

PR #165, opened 2016-08-01
Fix code analysis

hvelarde:
- Fix imports
- Do not use quotes on strings
- Do not use invokeFactory
- Do not depend on unittest2

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,18 @@ Changelog
 1.0.5 (unreleased)
 ------------------
 
+- Fix imports to follow conventions.
+  [hvelarde]
+
+- Avoid usage of double quotes on strings.
+  [hvelarde]
+
+- Avoid usage of invokeFactory.
+  [hvelarde]
+
+- Remove dependency on unittest2 as package is not intended to be compatible with Python 2.6.
+  [hvelarde]
+
 - Use selenium v2.53.6.
   [hvelarde]
 
--- a/bobtemplates/plone_addon/src/+package.namespace+/+package.name+/tests/test_+package.dexterity_type_name_lower+.py.bob
+++ b/bobtemplates/plone_addon/src/+package.namespace+/+package.name+/tests/test_+package.dexterity_type_name_lower+.py.bob
@@ -1,15 +1,14 @@
 # -*- coding: utf-8 -*-
-from plone.app.testing import TEST_USER_ID
-from zope.component import queryUtility
-from zope.component import createObject
-from plone.app.testing import setRoles
-from plone.dexterity.interfaces import IDexterityFTI
 from plone import api
-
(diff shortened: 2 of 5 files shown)


CHANGES.rst
```diff
@@ -4,6 +4,9 @@ Changelog
 1.0.5 (unreleased)
 ------------------
 
+- Fix code analysis.
```
tisto: @hvelarde could you please be a bit more specific here?

bobtemplates/plone_addon/src/+package.namespace+/+package.name+/tests/test_+package.dexterity_type_name_lower+.py.bob
```diff
@@ -1,15 +1,14 @@
...
-from plone import api
-
-from {{{ package.dottedname }}}.testing import {{{package.uppercasename}}}_INTEGRATION_TESTING  # noqa
 from {{{ package.dottedname }}}.interfaces import I{{{ package.dexterity_type_name }}}
+from {{{ package.dottedname }}}.testing import {{{package.uppercasename}}}_INTEGRATION_TESTING  # noqa
+from zope.component import createObject
+from zope.component import queryUtility
 
-import unittest2 as unittest
+import unittest
```
tisto: @hvelarde this means dropping Python 2.6 support on Plone 4. This should be reflected in the generated setup.py.

hvelarde: in fact, this was obviously a typo: Python 2.6 support was always broken because `unittest2` was never included as a dependency in `setup.py` and this is the only module that refers to it; I'm just reflecting that.

jensens: bobtemplates.plone generated packages do not have a troove classifier for Python 2.6, they state to be for 2.7:
https://github.com/plone/bobtemplates.plone/blob/master/bobtemplates/plone_addon/setup.py.bob#L26

jensens:
LGTM - Python 2.6 support was never a thing for bobtemplates.plone and it do not state to do so in its generated setup.py.

pbauer:
I use unittest2 in all my projects since it backports all features from python 3.5.
IMHO the most usefull features are the immensly improved error-messages and diffs when comparing dicts and such.

jensens:
Afaik all was already backported to Python 2.7 unittest.  What features specifically are missing?

jensens:
just because I'am curious whats the differences: looking at https://pypi.python.org/pypi/unittest2 section 'differences' shows, not much. So I suppose the removal of unittest2 is fine.

PR #166, opened 2016-08-02
Get rid of travis.cfg configuration as its use is no longer considered best practice

hvelarde:


bobtemplates/plone_addon/.travis.yml.bob
```diff
@@ -1,28 +1,26 @@
...
-  - echo "[buildout]" > $HOME/.buildout/default.cfg
-  - echo "download-cache = $HOME/buildout-cache/downloads" >> $HOME/.buildout/default.cfg
-  - echo "eggs-directory = $HOME/buildout-cache/eggs" >> $HOME/.buildout/default.cfg
+  - eggs
+matrix:
+  fast_finish: true
 install:
-  - python bootstrap-buildout.py -c travis.cfg
-  - bin/buildout -Nc travis.cfg
+  - python bootstrap-buildout.py
```
hvelarde: may I ask why we renamed the `bootstrap.py` script?

bobtemplates/plone_addon/.travis.yml.bob
```diff
@@ -1,28 +1,26 @@
 language: python
+python: 2.7
 sudo: false
 cache:
-  pip: true
   directories:
-    - $HOME/buildout-cache
```
tisto: @hvelarde why are you removing the buildout cache here? That's actually quite useful.

hvelarde: because is just not needed: we are already caching the `eggs` directory.

jensens: -1 - i just fixed the whole thing, but we need this. the buildout cache is not the pip cache.

hvelarde: I changing this only for the packages created with bobtemplates.plone, not for bobtemplates.plone itself; the `eggs` directory I'm caching is the one create by buildout.

why you want to keep that on the other packages? what problem are you solving?

bobtemplates/plone_addon/.travis.yml.bob
```diff
@@ -1,22 +1,26 @@
...
-python:
-  - 2.7
+  - eggs
```
hvelarde: @tisto @jensens the caching is done here.

bobtemplates/plone_addon/.travis.yml.bob
jensens: right, so touching the cache is out of scope, right? I spent quite some time to get this caching topic perfect. It reduces build time of test down to roughly 2 minutes.

-> caching the eggs directory is not enough, you want to cache downloads as well.

Also I simplified to pass this information about the caches explicit to buildout as you see on current master. Please sync the PR with it.

bobtemplates/plone_addon/.travis.yml.bob
jensens: May you point me to the related travis docs to make me understand the magic?
How does this cover download cache from buildout?

hvelarde: we list the directories we want to keep using the the `cache` option; buildout runs normally downloading and decompressing all eggs to the `eggs` directory in the root of the package; we include that directory in the cache; that's all; no magic, just plain buildout with no additional configuration.

https://docs.travis-ci.com/user/caching/#How-does-the-caching-work%3F

bobtemplates/plone_addon/.travis.yml.bob
hvelarde: no, the changes I made keep the cache also and Travis runs fast as you can see here:

https://travis-ci.org/collective/collective.fingerpointing
https://travis-ci.org/collective/collective.lazysizes
https://travis-ci.org/collective/collective.nitf
https://travis-ci.org/collective/collective.polls
https://travis-ci.org/collective/sc.social.like
https://travis-ci.org/collective/collective.upload

in some cases I also added the `parts/node` directory to the cache also.

bobtemplates/plone_addon/.travis.yml.bob
jensens: Ok, it indeed looks like the default ``eggs`` directory gets cached. 

But you do not catch the ``download-cache`` - which I would not like to loose. Download cache must be enabled explicit.

hvelarde: what's the default value for the `download-cache`?

jensens: you need to give it a value. Given the relative URL in the ``directories`` is the git clone path, this should do it:

```
directories:
  - buildout-cache

[...]

./bin/buildout [....] buildout:download-cache=${buildout:directory}/download-cache
```

jensens: this also means that nobody plays around with his/her own settings in buildout. therefore I tend to pass in the eggs-directory too, to be explicit.

hvelarde: IMO, this is not necessary as we already have the eggs cached, but… as you wish; let me make a few tests here.

jensens: This are complete different things, eggs downloads and the generic buildout download infrastructure.

jensens: Note: the former travis.cfg used the ``buildout.plonetest`` which downloads the Unified Installer. So caching downloads has there a large effect. 

If you do not use this any more the impact is smaller to zero. 

But it happens one later extends its buildout by something that downloads. I'd like the template offering this feature by default, and costs are almost zero.

bobtemplates/plone_addon/.travis.yml.bob
```diff
@@ -1,22 +1,28 @@
...
 python:
   - 2.7
+matrix:
+  fast_finish: true
 install:
-  - python bootstrap-buildout.py -c travis.cfg
-  - bin/buildout -Nc travis.cfg buildout:eggs-directory=$HOME/buildout-cache/eggs buildout:download-cache=$HOME/buildout-cache/downloads
+  - python bootstrap-buildout.py
+  - bin/buildout annotate
+  - bin/buildout buildout:download-cache=downloads
```
jensens: This should be (as it was before) ``bin/buildout -N buildout:download-cache=downloads`` to get most out of caching.

jensens: Btw., whats the ``annotate`` good for? I never got what's the value of it in a test environment. For local development it handy.

hvelarde: I use `annotate` for debugging purposes; sometimes you're not loading the right Plone version on your tests, or some specific package pinning and this is useful to debug that in CI.

hvelarde:
@tisto @jensens this is ready to be merged if you have no further comments.

jensens:
Sorry, I did not found time to really test the code in a live-travis configuration. I put this on my todo list.

jensens:
I just found time to test it. Works perfect and all is much simpler than before. I'll merge.

tisto:
@jensens travis is broken since you merged that pull request: 

https://travis-ci.org/plone/bobtemplates.plone/builds/163775407

Please fix or revert asap.

jensens:
@tisto well, I've been careful:  @hvelarde branch was green, it was rebased correctly. Looks (first, rough) more like some version pin was missing or some code analysis specific property changed meanwhile and that broke it. So revert would result in a broken master then as well. I'll look at it after weekend. Dont expect Armageddon because of a red master meanwhile.

tisto:
@jensens if you merge, it is your responsibility to check if the build it green afterwards. Fix or revert is our standard policy. No reason for unprofessional comments...

jensens:
@tisto In German we say "Wie man in den Wald ruft, so schallt es zurück". Nix für ungut ;)

hvelarde:
@tisto the problem is the flake8-coding, so @jensens is right; that has nothing to do with this PR.

there have been a lot of issues with code analysis recently because most plugins are now compatible with Flake8 3.0; I've been fighting with this for some time in the Collective.

BTW, plone.recipe.codeanalysis is not working fine locally for me... but that is another issue.

PR #171, opened 2016-09-21
include barceloneta resources, toolbar vars and Grunt setup for Themes

MrTango:
I included the barceloneta resources and add a Grunt setup for Theme add-ons.
This makes the start with the Theme packages a lot easier now.


bobtemplates/plone_addon/src/+package.namespace+/+package.name+/configure.zcml.bob
```diff
@@ -17,10 +17,29 @@
   <plone:static
       directory="theme"
       type="theme"
-      name="{{{ package.dottedname }}}"
+      name="{{{ package.name }}}"
+      />
+
+  <genericsetup:registerProfile
+      name="default"
+      title="Plone Theme: {{{ package.name }}}"
```
cdw9: Using just the `name` instead of `dottedname` here doesn't work so well if you have a different naming convention. I generally name themes `product.theme`, and in this case, the entry in the Add-ons panel just looked like "Plone Theme: theme".  I would recommend using `dottedname` here, like on the uninstall profile.

PR #174, opened 2016-10-09
WIP Update flake8 version

gforcada:
This probably fixes travis, although it probably will break some flake8 extensions that needs porting to flake8 >= 3

Let's see what travis says


tisto:
@gforcada I would like to merge that pr and do a release. I ran into that particular problem already...

PR #180, opened 2016-11-19
Stop xvfb after running scripts

gforcada:


tisto:
@gforcada why would we want to clean up? The VMs are destroyed after the run anyways. In my opinion this is unnecessary and just confused people.

gforcada:
@tisto note that this is on the bobtemplates.plone travis.yml not the generated one.

And so far the only one confused was travis itself. If you look at the previous master builds on travis all of them were hanging for almost an hour until travis killed them.

There was some unspecific Firefox debug message and this made it go away, so unless you have another way of fixing it, this works (TM) :smile:

PR #182, opened 2016-12-08
Fix violations on generated code. Add new documentation about the str…

sneridagh:
…ictness of plone.recipe.codeanalysis. Add .isort.cfg default config.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -134,6 +134,28 @@ Your buildout should be ready to use now.
 See `the documentation of mr.bob <http://mrbob.readthedocs.org/en/latest/>`_  for further information.
 
 
+Integration with plone.recipe.codeanalysis
+------------------------------------------
+
+This package is integrated by default with ``plone.recipe.codeanalysis`` that enforces the use of good known practices and the Plone code style guide.
+
+By default, the recipe setup a git precommit hook that performs the checks on the current package code. If it fins some violation the commit is not performed and the violations are shown.
+
+In order to fix all the violations, please refer to the plone.recipe.codeanalysis documentation: https://github.com/plone/plone.recipe.codeanalysis
+
+and https://github.com/plone/jenkins.plone.org/blob/master/docs/source/run-qa-on-package.rst
+
+The import sorting issues can be sorted out by installing the ``isort`` utility:
+
+.. code-block:: shell
+
+    pip install isort
+
+    isort src/**/*.py
+
+You can also install other utilities to help you to fix and avoid violations, like ``autopep8``.
+
+
 Contribute
 ----------
 
--- a/bobtemplates/plone_addon/.isort.cfg
+++ b/bobtemplates/plone_addon/.isort.cfg
@@ -1,0 +1,6 @@
+[settings]
+force_alphabetical_sort = True
+force_single_line = True
+lines_after_imports = 2
+line_length = 200
+not_skip = __init__.py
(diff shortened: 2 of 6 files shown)


README.rst
```diff
@@ -134,6 +134,28 @@ Your buildout should be ready to use now.
 See `the documentation of mr.bob <http://mrbob.readthedocs.org/en/latest/>`_  for further information.
 
 
+Integration with plone.recipe.codeanalysis
+------------------------------------------
+
+This package is integrated by default with ``plone.recipe.codeanalysis`` that enforces the use of good known practices and the Plone code style guide.
+
+By default, the recipe setup a git precommit hook that performs the checks on the current package code. If it fins some violation the commit is not performed and the violations are shown.
```
gforcada: That's not exactly true, even if there are violations the commit happens, you could force that though, but is not on by default (I  guess with the ``return-status-codes = True``).

bobtemplates/plone_addon/.isort.cfg
```diff
@@ -0,0 +1,6 @@
+[settings]
+force_alphabetical_sort = True
+force_single_line = True
+lines_after_imports = 2
+line_length = 200
+not_skip = __init__.py
```
gforcada: I would rather put these settings on ``setup.cfg`` (then the header needs to be changed to ``[isort]`` so we keep less files around and we have one single place to put all the configurations.

bobtemplates/plone_addon/src/+package.namespace+/+package.name+/tests/test_setup.py.bob
```diff
@@ -26,7 +26,7 @@ class TestSetup(unittest.TestCase):
         from {{{ package.dottedname }}}.interfaces import (
             I{{{ package.browserlayer }}})
         from plone.browserlayer import utils
-        self.assertIn(I{{{ package.browserlayer }}}, utils.registered_layers())
+        self.assertIn(I{{{ package.browserlayer }}}, utils.registered_layers())  #noqa
```
gforcada: *Totally opinionated*: if the problem is that the line can be too long then I would rather change it to:

```
self.assertIn(
    I{{{ package.browserlayer }}},
    utils.registered_layers(),
)
```

I would be very wary to add ``# noqa`` all over the place, because it defeats the purpose of adding QA in the first place :-)

bobtemplates/plone_addon/src/+package.namespace+/+package.name+/tests/test_setup.py.bob
```diff
@@ -48,4 +48,4 @@ class TestUninstall(unittest.TestCase):
         from {{{ package.dottedname }}}.interfaces import \
             I{{{ package.browserlayer }}}
         from plone.browserlayer import utils
-        self.assertNotIn(I{{{ package.browserlayer }}}, utils.registered_layers())
+        self.assertNotIn(I{{{ package.browserlayer }}}, utils.registered_layers())  # noqa
```
gforcada: ditto

README.rst
sneridagh: I double checked and you're absolutely right. Sorry I don't know how I looked into it before it seemed to me that the commit was not being applied because of the config of the pre-commit hook... in fact, the motivation for the whole PR was for fix it.

bobtemplates/plone_addon/.isort.cfg
sneridagh: The fact is that there are already there. T_T generate a command line script in ``bin`` for it would be great too.

bobtemplates/plone_addon/src/+package.namespace+/+package.name+/tests/test_setup.py.bob
sneridagh: Agreed.

bobtemplates/plone_addon/src/+package.namespace+/+package.name+/tests/test_setup.py.bob
sneridagh: I will close this PR and send a new one!

PR #185, opened 2016-12-23
Change default values for code analysis' return-status-codes directive

### LINKED ISSUE #184: bobtemplates plone breaks plone.recipe.codeanalysis pre-commit hook
Since we removed the travis.cfg and set "return-status-codes = True" in our buildout.cfg the p.r.codeanalysis pre-commit returns a status code 1. This prevents a commit to be executed if there is a code violation. This is extremely annoying for developers. It was a conscious design decision in p.r.codeanalysis to not prevent the commit but just notify the developer.

I would propose to either re-add a travis.cfg (quick fix) or find a better way to work around this (also fine with me but I don't have time to do it).

cc @hvelarde

hvelarde:
It is now False on development and True on CI.

closes #184

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,10 @@ Changelog
 1.0.6 (unreleased)
 ------------------
 
+- Change default values for code analysis' ``return-status-codes`` directive:
+  it is now ``False`` on development and ``True`` on CI.
+  [hvelarde]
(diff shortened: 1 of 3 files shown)


hvelarde:
I have no idea neither, maybe it doesn't like the `` chars.

PR #189, opened 2017-01-04
Update readme for buildout files settings

prakharjoshi:


prakharjoshi:
@gforcada Even for updating readme, do we need to check for changelog ?

gforcada:
@prakharjoshi you can safely ignore it :-)

PR #190, opened 2017-01-09
Sneridagh clarification backendxml

sneridagh:
Comment the toolbar rule by default in backend.xml and add a comment on how to add it properly if backend.xml is used. Declaring the toolbar rule twice causes the toolbar JS stop working.


sneridagh:
I've spent a good deal of time this evening figuring out what was happening when the toolbar JS stopped working for me :) Just making sure that doesn't happen again to anybody.

PR #195, opened 2017-02-17
[WIP] Fix Travis (fails on check-readme)

tisto:


.travis.yml
```diff
@@ -46,9 +46,6 @@ script:
   - bin/test
   - bin/test-plone_addon
 
-after_script:
```
tisto: @jensens could you elaborate why you think that xvfb might be the problem here? I don't see how the check-readme fail could be related. I'd be ok with just merging this pr to get back to a green master build. If we start to fiddle around in the same branch/pr, that becomes harder...

jensens:
sorry for abusing your PR, it is indeed out of scope of the title. I'll revert everything if I am on the wrong track.

Symptom is, that the build runs w/o errors and then it never stops and run in a timeout. I read somewhere that a hanging xvfb might be the reason and so I just tried this on your branch, since it's the only one w/o failure at the moment.

jensens:
I removed my experimental changes.

IMO we should merge.

The ``setup.py check `` is almost undocumented (or under documented) https://docs.python.org/2/distutils/apiref.html#module-distutils.command.check

tisto:
@jensens I do not want to merge my own pull request. I'm of course ok with merging it. :)

jensens:
@tisto Merged. The WIP in the title usually stops me from merging. Pro-tip: use the labels!

tisto:
@jensens right. my plan was to look into that later but I did not find the time...

PR #199, opened 2017-03-20
Spliting plone_theme_package template from plone_add, add plone_fattheme_buildout template

MrTango:
- splitting plone_theme_package template from plone_add, to make templates less complex and more testable
- add more tests for hooks and theme template
- add plone_fattheme_buildout template


tisto:
@MrTango sorry if I miss something here but would you mind elaborating why we need a buildout template for a theme? Wouldn't it be enough to just include that in the existing theme package? If a user wants to build a theme she would want the best practice in the buildout of the package anyways, right?

MrTango:
@tisto the buildout.cfg in the generated them package you mean right?
First we had this already in, i just splitted the theme from the addon, because it's getting more and more messi to have this in one template. So it should be separated templates, to make the templates not to complex. In the theme package we want to have also a development buildout like we have with the addon package. So that we can start Plone and develop and test the theme. Also we install the npm dependencies for the theme with buildout to make it simpler and ready to use after running the buildout.
This runs npm install in the buildout directory and also in the theme folder, where the theme dependencies are defined. This is separated, so that one is still able to work on the theme TTW. For example we don't want all resource like Grunt in the theme folder, because it is a lat, but we need npm packages like bootstrap & co inside the theme, so that the compiling TTW also works.

MrTango:
@pbauer please have a look at this PR, i would like to finish this soon, because updating it against the master is not so fun with the changes of the structure ;)
I'm also not so sure if the travis testing setup needs some more love here.

MrTango:
I also set the version to 2.0 because of the changes in the usage.
@pigeonflight was your part with the fat_theme_buildout (we need a better name) finished?
If you want to change some thing, please use a fresh checkout this branch, because i forced the push after rebasing master

MrTango:
@tisto the usage is documented in the theming training, only the new template not in there yet.
https://training.plone.org/5/theming/theme-package.html

tisto:
@MrTango is it possible that this new template is not tested at all? I do not see any changes in travis.yml or in the testing matrix:

https://github.com/plone/bobtemplates.plone/blob/Boston_theming_improments/.travis.yml#L21

When I cleaned up the package initially I removed lots of code and options to reach a point where we could actually test what we had. Our structure before was too complex to be testable and I fear that this is true for this pr as well. I would really like us to take testing seriously and think about how we can test this. 

I could try to help if I would understand the reasoning behind this. If the theme template differs too much from the other templates, it might be better to create a completely separate package and product (and test the theme package individually instead of trying to unify something that differs significantly).

MrTango:
@tisto I'm still working on the tests.
In fact because of the separated templates there is less to test, because we don't ask any more for the type, we just use the right template. But i also added a new question for the theme name and this is tested.
I also just added some more tests for some hooks, that where not tested so far.
Regarding the travis file, this could be improved a bit i gues, that's why I also wanted @pbauer to have a look at this before we merge the PR.

I'm against splitting the theming in a separate package, because it makes it more complex to use it for the users. One package with some useful templates, just like we had with ZopeSkel is what we need.

The plone_addon template, is also as useful as ZopeSkel was. We need a way to create sub commands here. But i think, this is possible as mrbob supports rendering multiple templates to the same target directory. So we could try to provide this. Especially for Dexterity CT's and all the boilerplate like the profiles stuff, this would be help people a lot. The next step after that, could be building a small cli-client which uses the mrbo api. So that we can actually have a cli like we have in other frameworks.

tisto:
@MrTango we currently have a two-dimensional testing matrix on Travis. If we add another top-level choice (and maybe even choices that only happen on one of the branches), we will end up with a (complex) three-dimensional matrix. 

Before you continue to put effort into this, please think about the testing matrix that your choices will introduce and try to come up with something simple (two dimensions, no tree, no choices that only occur on one of the branches).

I put a lot of effort into making this package testable and to reduce complexity. This made it possible to test what we have right now. We should learn from ZopeSkel and not make the same mistakes over and over again.

I am not against adding a more complex theme example. Though, I strongly oppose going back to the untestable state we had before the cleanup. In order to be testable we need a question structure that is simple and flat, otherwise, we will lose track of the things we test.

MrTango:
I agree, but the current situation is, that theming was never tested by it self.
Only the Basic or Dexterity paths including there questions/answers where tested.
This PR so far, add more tests to already existing parts, like theme specific answers, the hooks in general and a view new theme specific hooks.
So the improves the test coverage and makes the templates it self cleaner, because we get rid of some of the special conditions in the hooks, which where mainly to solve the problem that we used one templates for deferent things. I'm suggesting that we use Mrbob like it was thought and don't mix different things into one template. That makes things cleaner, less complex and better testable.
I'll also add more tests in the future, for example I'll also add separate answers to Travis.
In fact there already there, but we have to improve the Travis setup, so that it is testing every template with it's own questions.

tisto:
@MrTango I think you missed my point. Maybe I am missing yours. What about a quick phone call today or in the next days to fix that? :)

MrTango:
Ok, after talking to @tisto, here some notes and also ideas for the records:

We will split out the dexterity/basic functionalities into separate templates too, to get rid of the extra type-question end therefor rid of one dimension in the test matrix. We use the same dimension then for the different templates instead.

My ideas:

To improve the handling for user of the templates more, we should find a way to add additional pieces to existing projects generated by the templates. Basically this is supported by mrbob, which can apply multiple templates to the same folder structure. The most important thing to solve is, how we add things like a new content type and its needed configuration like ZCML and/or XML (GS profiles) safely.

So it would help already a lot if could do the following things for adding a content type:
_add a new bobtemplate for that which is doing:_

- create the type files in profiles/default/types/ct.xml
- maybe also add the include line in the profiles/default/types.xml file, should be possible to parse and add something here, it's XML 
- depending on the type of the model xml/schema based create the files with some example fields. 
    - structure: content/ct.py with schema interface or models/ct.xml with XML model definition
    - the same for other files like views/templates, so that we mostly create just files in folders, not extending files
    - exception could be ZCML configuration, but again it's XML so we can parse it and extend it, not need for stupid placeholders
- anytime we can't do some changes safely, we just print out what to do, so the user can do the steps

@tisto, @pbauer any opinions?

MrTango:
As the changes here doesn't make things worse, I'll merge it later.
Otherwise I'm afraid to have a lot of work, if some one make change to the master before.
We can slit out the Dexterity stuff later and add the new smaller templates for the content types.

gforcada:
Amother idea: for each extra that we add, always print at the end, regardless of if it could automatically modify or not all files, a link to a chapter on the docs/training where it explains especificcally that did that you just added.

Thus way users will have all the boilerplate needed and a reference to what all that means.

On the training/docs we could do the same (point to the mrbob template) so they can be mantained together.

Not sure if the docs should be here in this repo itself...

PR #206, opened 2017-07-06
Rework whole Package to work under Python 2+3 and tests

loechel:
Rework the whole Package to make it testable and follow best practices.
Used tox and pytest as new test environment.


loechel:
@MrTango / @gforcada ready for you.

tests/test_hooks.py
```diff
@@ -0,0 +1,274 @@
...
+    # Error Check
+    with pytest.raises(ValidationError, msg='Value must be a boolean (y/n)'):
+        assert hooks.to_boolean(None, None, 'spam')
+
+
+def test_validate_packagename():
+    with pytest.raises(AttributeError):
+        hooks.validate_packagename(None)
+
+    configurator = Configurator(
```
MrTango: Is here an indention missing?

loechel: no. that is like it should.

I do test various variants within one test method. 
As ``hooks.validate_packagename()`` should take a Configurator object but also tests against None did have for all of those case test steps. Maybe it would be good to put comments for each case.

loechel:
@MrTango commented to make it clearer

MrTango:
@loechel can you please add some docs. I also can not run the buildout, it crashes:

```
    return installer.install(specs, working_set)
  File "/home/maik/develop/plonecore/bobtemplates.plone/local/lib/python2.7/site-packages/zc/buildout/easy_install.py", line 708, in install
    extra_requirements = dist.requires(req.extras)[::-1]
  File "/home/maik/develop/plonecore/bobtemplates.plone/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2559, in requires
    "%s has no such extra feature %r" % (self, ext)
UnknownExtra: bobtemplates.plone 1.0.6.dev0 has no such extra feature 'test'
```

I also added a requirements.txt.

loechel:
@MrTango I added some documentation and removed more unnecessary files like buildout.cfg and requirements.txt 

You do not need any buildout on this package anymore it is fully pip compatible and should only be developed like that and also tested.

jensens:
If it is pip compatible I would expect a requirements.txt file...

loechel:
@jensens why,

A good python package should define all of it requirements in the setup.py even the constrains. requirements.txt and constains.txt are for python projects, where you plug several packages together.

So there is no need for a requirements.txt or buildout.cfg or anything project related here as it is a plugin for mr.bob and not a project itself.

jensens:
@loechel it is more about version pins, which should not happen (except minimum versions) in setup.[py|cfg]. So I agree not adding ``requirements.txt``, but if we don't use buildout, ``constraints.txt`` is a great place for this

README.rst
```diff
@@ -78,89 +78,101 @@ It is also available online at http://docs.plone.org/develop/addons/bobtemplates
...
-Or to create a new theme package::
+    # via pip - prefered way
+    $ pip install bobtemplates.plone
 
-    mrbob -O plonetheme.tango bobtemplates:plone_theme_package
-    cd plonetheme.tango/
-    pip install -r requirements.txt
-    buildout bootstrap
-    bin/buildout
+    # via easy_install - old way if pip is not avaliable
```
gforcada: What would be that case actually? We could link to pip documentation on how to install it if it's not available, but I would refrain of giving alternative routes...

MrTango:
@loechel can you also please add the docs for the developing part.
I mean usually we use zest.releaser to make a release, but this works in buildout, or is there enother way?
And what do I have to do after checking out the repo, to work on it?
Maybe add a how to develop section to the readme, which makes the way clear.
I'm open to new ways, if we have it clear.

MrTango:
We are also setting the precommit hook with buildout (codeanalyis) automtically and that should be done automatically. Is that possible without buildout.

Speaking about best practice, I'm not sure if we should try to change everything which is best practice for Plone so far. I mean I'm ok for smal changes like adding requierements.txt or constraints or what ever you call it. But removing buildout from the stack is a big step and so far not best practice and will bring a lot of things to solve in another way. We can discuss this more in detail and adopt more best practices from the Python world in time and with care to fullfill all the needs we have.

For this package, it might be ok to go without buildout, if we solve all tasks we had, but for sure not for the add_on template.

loechel:
@jensens I do disagree, I think those files: requirements.txt and constrains.txt did make sense for projects but not for a single package.

if you look into the current ``setup.py`` we do ´define just two dependencies:

    install_requires=[
        'setuptools',
        'mr.bob',
    ],

if we now would change that to also ensure constrains that would be like:

    install_requires=[
        'setuptools',
        'mr.bob >= 1.0, <= 2.0, != 1.2.3',
    ],

see: http://peak.telecommunity.com/DevCenter/setuptools#declaring-dependencies

for a broader discussion look at: https://packaging.python.org/discussions/install-requires-vs-requirements/

I would prefer to have those packages as small as possible and don't overload them with files that are used because most people don't know a better way.

loechel:
@MrTango I have added some documentation about developing and releasing. 

A git hooks are not automatically installable if the users / developers have a standard installation. You need to run a sort of script, either buildout or a shell script or do it manually. I personally prefer the manual way, but I also do remove the set pre-commit hook by plone.recipe.code-analysis, as this also overwrites all other pre-commit-hooks.

loechel:
About new best practices there is my PLIP: https://github.com/plone/Products.CMFPlone/issues/2072 and also my work on https://github.com/plone/plone_best_practices_discussion 

we should definitely discuss that at the Plone Conf to find a common sense, but we all need to move to make Plone and its development more approachable. 

But yes we need o think how to make a good skeleton for Plone core packages and Plone add-ons and both could differ.

loechel:
@MrTango about the best practice changes:

>Speaking about best practice, I'm not sure if we should try to change everything which is best practice for Plone so far. I mean I'm ok for smal changes like adding requierements.txt or constraints or what ever you call it. But removing buildout from the stack is a big step and so far not best practice and will bring a lot of things to solve in another way. We can discuss this more in detail and adopt more best practices from the Python world in time and with care to fullfill all the needs we have.

I do agree with you, we should not move to quickly for Plone Packages, but we also should not handle all packages in the Plone would like belonging to Plone Core or Plone itself. This Package bobtemplates.plone is not a package that you would ever install into a Zope Instance, so why should it follow Plone ways of doing. The defined skeletons inside or different things, they should follow Plone Best Practices and usage of buildout for some parts is best practice. As long as we do not move away form zope.testrunner we do need buildout to run our tests.

But general Python Packages should also adopt python best practices, and that at the moment is tox and pytest.

README.rst
loechel: fixed

loechel:
@tisto first of all, I would like you not to shut up and participate more in the discussion, your opinion is very well accepted and could help us all to move forward. 

second, this pull request was original a cleanup and enhancing test coverage pull request, that should enable another pull request that could merge the discussion about best practices and potential implementations for the PLIP into an umbrella. This pull request did not apply any new or discussed best practices in the skeletons.

The changes in the test setup and invocation here are just to enable that. That is by the way test the current implementation against several Python versions is a nice to have thing but completely to ignore.

I would also don't like to mix several discussions. So best practice discussions and discussion about usage of tox should go into the other repositories and not on this thread.

My goal would be that we could test this package as good as possible as well as test automatically all generated skeletons. If possible against all Plone Versions we try to support and that local and via CI Servers on the same way, so that test results are consistent.

If I did remove any additional tests, please point me on that and I will reintroduce them. 
I did not remove the test matrix, all tests should still be in place, just the invocation has changed and should be simplified.

.travis.yml
```diff
@@ -1,58 +1,51 @@
...
-  - mkdir $HOME/.buildout
-  - echo "[buildout]" > $HOME/.buildout/default.cfg
-  - echo "download-cache = $HOME/buildout-cache/downloads" >> $HOME/.buildout/default.cfg
-  - echo "eggs-directory = $HOME/buildout-cache/eggs" >> $HOME/.buildout/default.cfg
-  - pip install --upgrade pip setuptools zc.buildout
-  - if [[ `npm -v` != 3.10.7 ]]; then npm i -g npm@3.10.7; fi
-  - npm install -g grunt-cli
-
-env:
-  matrix:
```
tisto: @loechel this test matrix runs tests against multiple versions of Plone and for the different package types. It is a two-dimensional matrix and running those against different Python versions would mean we end up with a three-dimensional matrix. The tox environment only contains the Python version as a parameter. Therefore we do not run tests against multiple Plone versions and package types any longer as far as I can see.

loechel: Sorry @tisto but this way of running tests is extrem hard to understand on hard to extend. I have replaced that with a way that could solve that much more straight forward and still be readable.

To be in detail:
We should distinguish between this package itself (bobtemplates.plone) which should be tested, and all of its generated  packages. 
This package itself did not have any direct add-on relation to Plone, so testing against different Plone versions for this package makes no sense.
The generated packages, should be compatible with multiple Plone versions, so they need a useful test matrix to test against. Please look at https://github.com/plone/bobtemplates.plone/blob/tox/tests/test_skeleton_generation.py#L9-L83 for that.

My second point is, that tests should be runnable local and on CI without any difference, so I did move all the parts that are test related into the test-runner / test invocation part.

If all generated products would have have a tox.ini it would be much easier to test and maintain.

Any additional Feedback is welcome.

jensens:
I wonder if tox-travis https://pypi.python.org/pypi/tox-travis may help here

loechel:
@jensens previous, that would be a help, since travis-ci supports the TOXENV variable it is much easier and no additional plugin is necessary. 

See: https://github.com/plone/plone.api/blob/master/.travis.yml and https://github.com/plone/plone.api/blob/master/tox.ini as an example. 
This is also a thing that I would like to automate with the best practices, but that is not focus of this pull request.

.travis.yml
tisto: @loechel you did not "replace" the testing of the different package types and Plone versions. You just removed it altogether.

loechel: @tisto ok, please have a look again at the file https://github.com/plone/bobtemplates.plone/blob/tox/tests/test_skeleton_generation.py I have changed that so that it reflects your test matrix.

loechel:
Ok I have now moved the test matrix into the travis-ci config again. 
It tests now all. but the fattheme test itself fail, could someone else please have a look.

.travis.yml
tisto: @loechel the link does not work. As far as I can see we still do not test the different package types. I would also keep in mind the overall execution time of the full test matrix. You usually don't want to run this at all on your local machine.

As I told you during our discussion at the Midsummer sprint, I have some fundamental doubts about using tox on a CI system. There is a fundamental mismatch between a local command like tox and CI which can not be overcome easily (see for instance this tweet from one of the pytest coredevs https://twitter.com/ossronny/status/894442880789557249).

I would suggest that we figure out where we want to go, before putting too much effort into discussions about implementation details. The PLIP is the right place for this discussion, not bobtemplates.plone.

loechel: @tisto sorry I have worked on several things in the mean time and moved files around, so that those links did not work anymore.

Travis currently shows the best information about status, and reflects that we do test the different package types, (well I have tested with all, and excluded the Plone_fattheme_buildout for now, as those seems o be totally brocken and tests always fail) 

Yes I did hat problems with the execution time, the tests has needed more than 2 hours locally and on CI on the first runs of all tests, we are now down to less then 20 Minutes.

I do share your doubts about using tox on CI, it is an unnecessary layer that could make more confusion than actual improvement. The problem is, that CI tools like Travis-CI did not allow that complex configuration that is possible with tools like buildout or tox. So you need an abstraction layer in between to make complex test setups work. 

My Suggestion would be, that we do a larger discussion at the PloneConf, maybe an open space and discuss how to continue. 
As far as I am now, I would work on documentation of my ideas for better best practices including tox, on a separate repository: https://github.com/plone/plone_best_practices_discussion instead of the suggested way of doing that as a pull request on bobtemplates.plone 

This pull request isn't about using tox or things like that, even if I originally started with this intend. This packages was in my opinion in such a bad shape, that it need a cleanup and more tests, so that additionally pull requests could focus on documentation, features, or refactoring. And that is what this pull is about, I did not touch any functionality, or the skeletons beside formatting for readability and making them testable and fixing test errors. 
And a cleanup consists of removing harmful elements.

loechel:
this pull gets partial superseeded by #216

loechel:
this pull gets partial superseeded by #216

loechel:
most of the stuff from this branch has already been merged into 3.0. so I close this branch.

PR #207, opened 2017-08-02
Set the zope_i18n_compile_mo_files environment variable.

mauritsvanrees:
Otherwise no `.mo` files are created when Plone starts up and you do not get any translations.


loechel:
OK the tests on master are almost useless and fail on pity things like version conflicts, so I do merge as it they fail.

PR #217, opened 2017-09-25
3.0

loechel:


MrTango:
Merging refactoring, to have subtemplates, pytest and tox available

CHANGES.rst
```diff
@@ -5,6 +5,17 @@ Changelog
 ------------------
 
 - Nothing changed yet.
+- Removed bootstrap-buildout.py and added DEVELOP.rst
```
idgserpro: https://github.com/plone/Products.CMFPlone/issues/2001 is still opened. Should we really remove bootstrap from here?

loechel: @idgserpro this is not a Zope/Plone element package, it is a pure python package / mr.bob add on which generates skeletons for Plone, so this PLIP did not apply here. 

And yes it is a good idea to remove elements that could confuse users and developers.

idgserpro: The PLIP is not from a python package but it's decision inflicts change in behavior here, that's why the questioning. Since bootstrap.py is not officially removed from Plone and we don't have the following deliverables:

> Pull requests to documentation repository with:

> the bootstrap to virtualenv update
> the porting guide
> a basic introduction to virtualenv

We just think this should be settled down before changing everything into a module that's supposed to be the starting point of official Plone practices when creating an addon: in other words, since this decision is not *finished* it shouldn't be forced to new addons.

This is just my 0.02c here, [other packages](https://github.com/reinout/z3c.dependencychecker/pull/45) are not concerned about this.

loechel: @idgserpro I am sorry if I did not get it right, but all the generated skeletons still contains a bootstrap.py so that plone/Products.CMFPlone#2001 is still valid for those generated addons. That is nothing we wanted to change with this 3.0 Version.

What we changed and removed was the bootstrap.py for bobtemplates.plone itself, as this is not a Plone core or addon package in the sense: ``Framework :: Plone`` Depending Package.

idgserpro: It makes sense. Anyways, thanks for your work on the package.

PR #224, opened 2017-10-08
fix defaut travis setup for new addon and theme packages

### LINKED ISSUE #222: default travis setup is broken
```
$ python bootstrap-buildout.py

python: can't open file 'bootstrap-buildout.py': [Errno 2] No such file or directory
```

pbauer:
Fixes #222

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,10 +4,12 @@ Changelog
 3.0.0a3 (unreleased)
 --------------------
 
-- Nothing changed yet.
+- Fix travis-setup for created packages wihtout bootstrap.py (Fixes #222)
+  [pbauer]
 
 
 3.0.0a2 (2017-10-01)
+--------------------
 
 - Cleanup Package - remove unnecessary files from past versions
   [loechel]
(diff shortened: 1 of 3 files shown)


jensens:
Your change is an outdated way to cache stuff, (I came up at some point with this overly complex style). 

@hvelarde came up with a much simpler approach I like more. 
Also avoiding bootstrap.py is overall a very good idea as well.

At current projects the ``.travis.yml`` looks like here so, and its very fast and w/o problems. 
I just have no idea if the ``requirements.txt`` from 5.1 plays well with a Plone 4.3-latest. 
At least here, it works with 5.0-latest.
I would prefer to have the ``requirements.txt`` placed at http://dist.plone.org, but this has not happened so far.

```
...
cache:
  pip: true
  directories:
  - eggs
  - downloads
env:
  matrix:
    - PLONE_VERSION=5.1
...
install:
  - sed -ie "s#test-5.1.x#test-$PLONE_VERSION.x#" buildout.cfg
  - pip install -U pip
  - pip install -r https://raw.githubusercontent.com/plone/buildout.coredev/5.1/requirements.txt
  - buildout -N buildout:download-cache=downloads code-analysis:return-status-codes=True annotate
  - buildout -N buildout:download-cache=downloads code-analysis:return-status-codes=True
...
```

PR #225, opened 2017-10-09
fixed syntax error in theme package grunt template

sherpya:


pbauer:
@sherpya thanks for the constribution! I added the same fix it to #229 to not start discussions about the Contributors Agreement 😃

PR #230, opened 2017-10-20
make sure to use the newest python version 

pbauer:
the one in virtualenv is too old for Plone 5.1


jensens:
hmm, strange, just declaring
```
python:
  - 2.7.14
```
does not help?

pbauer:
No. The command `virtualenv` uses a different python by default.

jensens:
thats odd

MrTango:
I solved this in EasyNewsletter by  doing "pip install --upgrade virtualenv" in before_install section

PR #232, opened 2017-10-23
Add missing parameter to make subtemplates work

erral:
`configurator` parameter is missing from the call and breaks when you try to call a subtemplate


erral:
I close this, it's already fixed on Master.

PR #233, opened 2017-10-23
Use uppercased content-type names

erral:
I think that to be consistent with the basic content-types provided by Plone, the ones created using bobtemplates.plone should also be uppercased (`Video` instead of `video`).

What do you think?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changelog
 3.0.0a3 (unreleased)
 --------------------
 
+- Create uppercased content-types.
+  [erral]
+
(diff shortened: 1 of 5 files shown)


pbauer:
I usually create owercase contenttypes and would preper them. We kept the default uppercase in plone.app.contenttypes for compatability and to ease migrations. In the training (e.g. https://training.plone.org/5/mastering-plone/dexterity_3.html) we create lower-case types.

erral:
Yes, I saw that in the training. That's why I was asking the difference between the default ones in Plone and the new ones :)

PR #236, opened 2017-11-07
Fixes #235: Different roles in tests for install and uninstall

### LINKED ISSUE #235: Different roles in tests for install and uninstall.
By default the ``test_setup.py.bob`` defines two tests ``TestSetup`` and ``TestUninstall``. 

In ``TestSetup`` everything was already executed because this happens in the test layers. The user there has the ``Manager`` role. The test just checks if everything is there.

In ``TestUninstall`` the uninstall is executed in the test with default roles of the test, that is the ``Member`` role. 
In my opinion, this is wrong. Before uninstall, the Manager role has to be adopted.

I stumbled across this because I wanted to remove an ``OFS.folder.Folder`` in setuphandlers ``uninstall`` I created at ``post_install``.

I propose to add a ``setRoles(self.portal, TEST_USER_ID, ['Manager'])`` to ``TestUninstall`` prior to the ``uninstallProducts`` call.

Opinions?

jensens:


CHANGES.rst
```diff
@@ -4,7 +4,9 @@ Changelog
 3.0.0a4 (unreleased)
 --------------------
 
-- Nothing changed yet.
+- Fixes #235: Different roles in tests for install and uninstall.
+  Now both is executed as role ``Manager``.
```
mauritsvanrees: This confuses me. Are the roles different or are they both Manager? I guess I see what you mean. Maybe change it to this:

Use Manager role in uninstall test.
Then both install and uninstall are executed with role Manager.

mauritsvanrees:
The linting tests on Travis fail, which seems unrelated to this issue.

MrTango:
This PR broke the build!
It was red before but that was only the linting part, now also some skeleton test are broken.
Please do not merge this kind of stuff.

MrTango:
https://travis-ci.org/plone/bobtemplates.plone/jobs/303610862#L5951

mauritsvanrees:
I am pretty sure I only saw errors in the linting tests, and no others.
But Travis says I am wrong: https://travis-ci.org/plone/bobtemplates.plone/builds/299127793

```
NameError: global name 'TEST_USER_ID' is not defined
```

Seems easy enough to fix at first glance. But we may want to revert the merge for now and get it green first.

jensens:
Yes, there were only error in the linting tests before, I double checked this.

MrTango:
It would be good if linting errors would have a different color like on jenkins ;)

MrTango:
@mauritsvanrees I have fixed this already locally. Don't worry.

PR #237, opened 2017-11-08
Fix coveralls by converting .coverage from pickle to json

pbauer:
Not sure if this is the current best practice. See https://community.plone.org/t/createcoverage-coveralls-io-version-conflicts/1188

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,8 @@ Changelog
 3.0.0a4 (unreleased)
 --------------------
 
-- Nothing changed yet.
+- Fix coveralls for packages created with addon and theme_package by converting the pickle output of createcoverage in .coverage to json.
+  [pbauer]
 
 
 3.0.0a3 (2017-10-30)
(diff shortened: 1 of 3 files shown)


pbauer:
At least this works for https://github.com/collective/collective.fieldedit. See https://coveralls.io/github/collective/collective.fieldedit

hvelarde:
sorry for the delay; I don't see the point on merging this: what was the problem you are trying to solve?

jensens:
The point is, the pickle format created by the version of coverage used by Plone by default do not store json but pickles, but coveralls expect json. 
Sorry, after all it works for me and others I now merged it to at least have some solution. Proposals for better solutions or new PRs welcome.

hvelarde:
I don't understand: I have this installed and working in literally dozens of add-ons based on Plone default configuration which uses `coverage = 4.2` as stated above; see for instance `collective.cover` as mentioned before by @mauritsvanrees:

* https://github.com/collective/collective.cover/blob/master/.travis.yml#L38-L41
* https://coveralls.io/github/collective/collective.cover

PR #244, opened 2017-12-06
Update setuptools and zc.buildout versions

tmassman:
Use newest versions for requirements.txt files and unpin versions in buildout configurations.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,9 @@ Changelog
 - fix linting problems and error in theme_package tests
   [MrTango]
 
+- Update setuptools and zc.buildout versions.
+  [tmassman]
+
 
 3.0.0a3 (2017-10-30)
 --------------------
--- a/bobtemplates/plone/addon/buildout.cfg.bob
+++ b/bobtemplates/plone/addon/buildout.cfg.bob
@@ -80,7 +80,7 @@ plone.app.dexterity = 2.1.1
 {{% endif %}}
 
 # development dependencies (tools and pre commit hook)
-setuptools = 33.1.1
+setuptools =
 zc.buildout =
 PyYAML = 3.12
 argh = 0.26.2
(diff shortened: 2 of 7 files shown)


gforcada:
@tmassman what's the idea of unpining on the cfg files? They should rather keep the version numbers in sync with requirements.txt IMHO.

tmassman:
@gforcada, well first to make it consistent. In some files they have been pinned, in some not. @MrTango started this [in his last pull request](https://github.com/plone/bobtemplates.plone/pull/241/files) already.
Second, it is way better to use the versions installed with pip into the virtualenv and refer to that from within buildout.

gforcada:
@tmassman so you mean that given we use a virtualenv, not pinning them rather than trying to get the lastest version they will use the one provided by virtualenv?

Stil, IMHO, we should pin them, because every now and then, one forgets to enable a virtualenv, or doesn't know about it (this is bobtemplates.plone, meant to help newcomers as well as seasoned, but newcomers are the ones that will look more for this kind of package generator).

And finally I always add ``allow-picked-versions = false`` to my buildouts, better to fail early and report that there are unpinned versions rather than 2 months later start getting errors because an unpinned dependency got bumped and now all goes astray.

MrTango:
I thing the way how you install it nowerdays, is first ``pip install -r requirements.txt`` and this way you have the correct versions anyway. There is no bootstrap.py anymore, which can give you the wrong versions.
@gforcada I guess the allow-picked-versions = false conflicts with the unset version, so we have to set it in this case. I personally don't use this. Buildout is much better now in showing what is wrong with versions and requirements ;).
For the bobtemplate, maybe we can put the versions for these into a separate place and write them into the templates, so that we don't have to change them in two places all the time.

MrTango:
I merged this, to have the newer version in.
The pinning we can add later with the parameter way.

PR #246, opened 2017-12-18
Improve documentation template, generate valid reST

svx:
This PR improves the bb template for documentation.

By adding:
``{{% for i in package.dottedname %}}={{% endfor %}}``

We make sure that ``=`` is always the same length as the ``package.dottedname``.

Doing this makes sure we generate valid reST !

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,8 @@ Changelog
 - Update setuptools and zc.buildout versions.
   [tmassman]
 
+- Imporve docs template for valid reST generation
+  [svx]
 
 3.0.0a3 (2017-10-30)
 --------------------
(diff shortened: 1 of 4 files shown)


tmassman:
I just saw there are more places where we could fix that, e.g. in the theme_package template. Could you do this as well?

svx:
@tmassman Ok I did for a few !

In general there is more work needed :)

Lots of the reST templates could see some love, we have there reST syntax issues and also the style, tone, voice, etc should see some attention some day :)

tmassman:
I created a new issue #247 for the updates on the doc.

Thanks again for your help on that topic, @svx!

MrTango:
@svx a test would make sense, to make sure we have valid reST.
Could you please provide one, so that we test all generated packages agains it?

PR #248, opened 2018-01-10
Pigeonflight fix pypi index

pigeonflight:
Fixed the pypi index to explicitly use https://pypi.python.org/simple/
For add-ons and theme packages.


pigeonflight:
@loechel appveyor is breaking my pull request :(

PR #279, opened 2018-04-20
[WIP]: test to include skeleton coverage

loechel:


setup.cfg
```diff
@@ -42,7 +42,11 @@ isort_ignore =
 
 [coverage:run]
 branch = True
-source = bobtemplates
+source =
+    bobtemplates
+    collective.foo
```
MrTango: This folder shouldn't exist actually. I tried to remove any trigger which might create the test packages directly inside the package. What we need to include here are the skeleton test which are created in a temporary folder.

something like:

`/tmp/pytest-of-maik/pytest-135/test_behavior0/collective.task/`

maybe we can set the temp directory for pytest to something else if needed.
If this approach works, it might probably make sense to have one test which include all sub-templates.

loechel: I totally agree those folders should not exists or life in the repository.
I will look and try to remove those creating calls, if possible.

For the skeleton test coverage, I am not sure if that is that easy.

jensens:
There was no activity on this for a while. I close this request for now. If you plan to work further on this please reopen and go on.

Issue #274, opened 2018-04-04
If global_allow is False, ask for parent containter content_type


MrTango:
If the user choose dexterity_type_global_allow: False we should ask for the conter content_type name we want to make the content type addable. This will allow us to configure the allowed_content_types setting in the FTI of the parent container content type and also generate working tests which uses the right container objects to create an content_type in. 

The mrbob question should named as follow:

`dexterity_parent_container_type_name`

MrTango:
to create the parent container, we can use the constructContent method from portal_types tool, which ignores the FTI settings like global_allow:

```
portal_types = self.portal.portal_types
parent_id = portal_types.constructContent(
    'my_parent_container_type',
    self.portal,
    'parent_id',
    title='Parent container',
)
self.parent = self.portal[parent_id]
```

This way we only care about the settings of the CT we are currently checking.
To create the curent CT object, we are using the plone.api, which will also respect the settings for our CT.

MrTango:
the settings global_allow and allowed_content_types are made in the FTI settings, which are defined in /profiles/default/types/<MyContenttype>.xml

MrTango:
To update the xml file, you have to add a method like "_update_metadata_xml" but name it "_update_ct_types_fti_xml". in this method you add the curent CT name to the list of allowed_content_types

MrTango:
set filter_content_types to True for the parent container, so that the parent container will actually filter the ctonent types which are alllowedf inside this container.

MrTango:
I created a small issue #286 for the Item content type, so that we don't have the filter options for an Item.

MrTango:
As discussed, we also need to update the robot test, depending on the global_allow and allowed_content_types settings. We need to use a Python based robot keyword, which will create the parent container, like we are doing in the unit tests already.

kakshay21:
Closing it now. If I missed anything, we can reopen it

PR #284, opened 2018-05-09
Add parent container in content_type

kakshay21:
see #274

### DIFF
--- a/bobtemplates/plone/content_type.py
+++ b/bobtemplates/plone/content_type.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+"""Generate content type."""
 
 from bobtemplates.plone.base import base_prepare_renderer
 from bobtemplates.plone.base import git_commit
@@ -14,11 +15,13 @@ import re
 
 
 def is_container(configurator, question):
+    """Test if base class is a container."""
     if configurator.variables['dexterity_type_base_class'] != 'Container':
         raise SkipQuestion(u'Is not a Container, so we skip filter question.')
 
 
 def check_dexterity_type_name(configurator, question, answer):
+    """Test if type name is valid."""
     if keyword.iskeyword(answer):
         raise ValidationError(u'{key} is a reserved Python keyword'.format(key=answer))  # NOQA: E501
     if not re.match('[_a-zA-Z ]*$', answer):
@@ -26,9 +29,14 @@ def check_dexterity_type_name(configurator, question, answer):
     return answer
 
 
+def check_global_allow(configurator, answer):
+    """Skip parent container name if global_allow is true."""
+    if configurator.variables.get('dexterity_type_global_allow', False):
+        raise SkipQuestion(u'global_allow is true, so we skip parent container name question.')  # NOQA: E501
+
+
 def _update_metadata_xml(configurator):
-    """Add plone.app.dexterity dependency metadata.xml in Generic Setup
-    profiles."""
+    """Add plone.app.dexterity dependency metadata.xml in Generic Setup profiles."""  # NOQA: E501
     metadata_file_name = u'metadata.xml'
     metadata_file_dir = u'profiles/default'
     metadata_file_path = configurator.variables['package_folder'] + '/' + \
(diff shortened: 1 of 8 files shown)


bobtemplates/plone/content_type.py
```diff
@@ -26,6 +26,11 @@ def check_dexterity_type_name(configurator, question, answer):
     return answer
 
 
+def check_global_allow(configurator, answer):
+    if configurator.variables['dexterity_type_global_allow']:
+        raise SkipQuestion(u'global_allow is false, so we skip container name question.')
```
loechel: this raises a flake8 error because the line is to long, 
please test such things by using tox upfront to check coding conventions are meet

loechel:
Looking better, could you add a test for that code, please

kakshay21:
sorry typo in the commit message that was a skeleton test.

bobtemplates/plone/content_type.py
```diff
@@ -143,6 +151,46 @@ def _update_rolemap_xml(configurator):
     update_file(configurator, file_path, match_str, insert_str)
 
 
+def _update_ct_types_fti_xml(configurator):
+    if (
+        (not configurator.variables.get('dexterity_type_global_allow', False)) and   # NOQA: E501
+        configurator.variables.get('dexterity_parent_container_type_name', False)    # NOQA: E501
+    ):
```
MrTango: i think the brackets for the if statement are not needed here.

bobtemplates/plone/content_type.py
```diff
@@ -143,6 +151,46 @@ def _update_rolemap_xml(configurator):
...
+        )
+
+        with open(file_path, 'r') as xml_file:
+            parser = etree.XMLParser(remove_blank_text=True)
+            tree = etree.parse(xml_file, parser)
+            tree_root = tree.getroot()
+            permname = """<element value="{0}" />""".format(
+                configurator.variables['dexterity_type_name'],
+            )
+            xpath_selector = ".//permission[@name='{0}']".format(permname)
```
MrTango: why are you searching for a permission tag here?

MrTango: please write a package-test, which tries to update a file, which has already the correct element in the list.
This will proably fail by now, so you can fix it then.

bobtemplates/plone/content_type/content/+dexterity_type_name_normalized+.py.bob
```diff
@@ -1,4 +1,7 @@
 # -*- coding: utf-8 -*-
+{{% if not dexterity_type_global_allow %}}
+from plone import api
```
MrTango: this is probably a left over ;)

skeleton-tests/test_addon_content_type.py
```diff
@@ -142,6 +142,52 @@ def test_addon_content_type(tmpdir, capsys, config):
...
+            '--config', 'answers.ini',
+            '--non-interactive',
+        ],
+        cwd=wd,
+    )
+    assert result == 0
+
+    # generate subtemplate content_type with child container:
+    template = """[variables]
+dexterity_type_name=parent
```
MrTango: parent again?
You have to create the children here and make the correct settings for it, as you would test it manually.

bobtemplates/plone/content_type.py
```diff
@@ -143,6 +151,46 @@ def _update_rolemap_xml(configurator):
     update_file(configurator, file_path, match_str, insert_str)
 
 
+def _update_ct_types_fti_xml(configurator):
+    if (
+        (not configurator.variables.get('dexterity_type_global_allow', False)) and   # NOQA: E501
```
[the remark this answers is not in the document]
kakshay21: this method only deals with dexterity_type_global_allow == False. If you see below it actually edits parent xml file once it's false. For true case this doesn't have to edit anything as things are already in place for true case.

bobtemplates/plone/content_type.py
kakshay21: you mean between two `and` case? Save check 😅

bobtemplates/plone/content_type/content/+dexterity_type_name_normalized+.py.bob
kakshay21: yeah 😅

bobtemplates/plone/content_type.py
MrTango: I mean:

`if not configurator.variables.get('dexterity_type_global_allow', False) and configurator.variables.get('dexterity_parent_container_type_name', False):`

and maybe better write it in a variable before:
parent_ct_type_name = configurator.variables.get('dexterity_parent_container_type_name', False) to make it shorter.

MrTango: also:
`configurator.variables.get('dexterity_type_global_allow') `
instead of:
`configurator.variables.get('dexterity_type_global_allow', False)`

get will return None in case there is not setting, which is fine.

kakshay21: On it!

package-tests/test_content_type.py
```diff
@@ -46,7 +49,79 @@ def test_prepare_renderer():
...
+            'non_interactive': True,
+        },
+        variables={
+            'dexterity_type_name': 'child',
+            'dexterity_type_global_allow': 'f',
+            'dexterity_parent_container_type_name': 'parent',
+        },
+    )
+    configurator.variables['package_folder'] = package_path
+    content_type._update_ct_types_fti_xml(configurator)
```
MrTango: let's check the generated xml here, so that we are sure that the changes are made correctly.
This will also improve the test coverage.

Issue #212, opened 2017-09-01
Prepare package for mockup pattern development


tomgross:
To promote mockup pattern development this package should include some parts to get started with mockup development.

Minimal:
1. Add compile-resources script to buildout

```
[zopepy]
recipe = zc.recipe.egg
eggs =
   Products.CMFPlone
   ${instance:eggs}
   ${test:eggs}
interpreter = zopepy
scripts =
   zopepy
   plone-compile-resources
```

2. Extend gitignore:

```
Gruntfile.js
node_modules/
package.json
```

MrTango:
We could create a subtemplate which adds a mockup pattern and also enables the tooling.
Could you try to add this in a branch?
I can help you in integrating this in bobtemplates.plone, but it's not that hard anymore.

MrTango:
The subtemplate should update the buildout file generated by the addon template.

MrTango:
@tomgross one question, now the script is only in the buildout template. Do you want it in the addon template too or instead?

tomgross:
I'd put it into the addon template only.

kakshay21:
I'll do that tomorrow

PR #291, opened 2018-05-28
Add mockup pattern development support in buildout

kakshay21:
see #212

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,8 @@ Changelog
 3.3.1 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Adding plone-compile-resources script in the buildout and extending .gitignore file
+  [kakshay21]
 
 
 3.3.0 (2018-05-24)
--- a/bobtemplates/plone/buildout/buildout.cfg.bob
+++ b/bobtemplates/plone/buildout/buildout.cfg.bob
@@ -15,7 +15,9 @@ always-checkout = true
 versions = versions
 extensions = mr.developer
 
-parts += test
+parts = 
+    test
+    plone-helper-scripts
 
 
 eggs +=
@@ -43,3 +45,13 @@ defaults = ['--auto-color', '--auto-progress']
 eggs =
     Plone
     ${buildout:eggs}
+
+[plone-helper-scripts]
+recipe = zc.recipe.egg
+eggs =
+   Products.CMFPlone
+   ${instance:eggs}
+interpreter = zopepy
+scripts =
+   zopepy
+   plone-compile-resources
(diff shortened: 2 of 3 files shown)


bobtemplates/plone/buildout/buildout.cfg.bob
```diff
@@ -42,4 +42,15 @@ recipe = zc.recipe.testrunner
...
-    ${buildout:eggs}
\ No newline at end of file
+    ${buildout:eggs}
+
+[zopepy]
+recipe = zc.recipe.egg
+eggs =
+   Products.CMFPlone
+   ${instance:eggs}
+   ${test:eggs}
```
jensens: I don't think we need the test-eggs here

bobtemplates/plone/buildout/buildout.cfg.bob
```diff
@@ -42,4 +42,15 @@ recipe = zc.recipe.testrunner
 defaults = ['--auto-color', '--auto-progress']
 eggs =
     Plone
-    ${buildout:eggs}
\ No newline at end of file
+    ${buildout:eggs}
+
+[zopepy]
```
jensens: In order to reduce confusion, I would name either name the section ``plone-helper-scripts`` or similar.

MrTango: regarding the zopepy it is useful for using autocompletion in editors like sublime text or vs code.
If you set the Python interpreter for the editor, your editor will have all buildout packages and do autocompletion. @jensens if you know a better generic way to do archive this, i'm happy to here it.

bobtemplates/plone/buildout/buildout.cfg.bob
MrTango: for autocompletion test-eggs meight be usefull too, so i would keep it in here.

CHANGES.rst
```diff
@@ -4,7 +4,8 @@ Changelog
 3.3.1 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Add mockup pattern development support in buildout
```
MrTango: Please descibe what you actually did, not what the title of the issue was.
Think about what yu PR will add and describe it.

bobtemplates/plone/buildout/buildout.cfg.bob
MrTango: naming the part: plone-helper-scripts is a good idea, let's do it.

bobtemplates/plone/buildout/buildout.cfg.bob
kakshay21: Do we need test-eggs then? @jensens @MrTango

PR #293, opened 2018-06-06
Update CHANGES.rst

kakshay21:


bobtemplates/plone/base.py
```diff
@@ -24,10 +24,14 @@
 
 
 def to_boolean(value):
+    """ A wrapper for mrbob.hooks.to_boolean
```
MrTango: unnecessary doc string, one can easily see what the method does ;)
And please write the if conition the other way around, to avoid the nesting of the call.
```
if not value:
    return
```

Issue #45, opened 2015-02-21
Add buildout script to create/update locales.


tisto:
See https://github.com/plone/bobtemplates.plone/issues/42

cedricmessiant:
There is an update.sh script which uses i18ndude in locales directory : https://github.com/plone/bobtemplates.plone/blob/master/bobtemplates/plone_addon/src/%2Bpackage.namespace%2B/%2Bpackage.name%2B/locales/update.sh.bob

frisi:
this script is a good start.
i'd add a shebang.

also, i'm wondering where $domain comes from.
isn't a  `domain = {{{ package.dottedname }}}` necessary?

also, for the manual translations for the plone domain, i'd write plone directly instead of re-assigning $domain


```
#!/bin/bash
# i18ndude should be available in current $PATH (eg by running
# ``export PATH=$PATH:$BUILDOUT_DIR/bin`` when i18ndude is located in your buildout's bin directory)

domain={{{ package.dottedname }}}
i18ndude rebuild-pot --pot $domain.pot --create $domain --merge $domain-manual.pot ../
i18ndude sync --pot $domain.pot */LC_MESSAGES/$domain.po

i18ndude sync --pot plone-manual.pot */LC_MESSAGES/plone.po
```

we should also document somehow how one can add translations for a language?
is it necessary to create the directroy manually? `mkdir de/LC_MESSAGES/`
and everything else is done by `./update.sh`?
or does on need to init the po file manually too `msginit --locale=en --input=../../$domain.pot`?

PR #294, opened 2018-06-09
locale update/create 

kakshay21:
see #45

### DIFF
--- a/bobtemplates/plone/addon/buildout.cfg.bob
+++ b/bobtemplates/plone/addon/buildout.cfg.bob
@@ -15,6 +15,7 @@ parts =
     i18ndude
     omelette
     robot
+    plone-helper-scripts
 develop = .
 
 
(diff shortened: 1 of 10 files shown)


bobtemplates/plone/addon/buildout.cfg.bob
```diff
@@ -91,6 +91,12 @@ eggs = zest.releaser
 recipe = zc.recipe.egg
 eggs = i18ndude
 
+
+[translate]
```
MrTango: please pay a bit more attention on how you name things!
What does this part do?
Also this is the name which will be placed into the bin folder.
So we want something there, which makes sense.
And please generate a package and test if it works ;)

bobtemplates/plone/addon/buildout.cfg.bob
```diff
@@ -91,6 +91,12 @@ eggs = zest.releaser
...
+recipe = plone.recipe.command
```
MrTango: We have a problem with this recipe. It only runs when we run buildout. Thats not what we want here. We need to create a script in bin folder. I'll copy my comments from the chat here too, for documentation:

We could use Python entry points for the locales script. We can't use a shell script, because we want also support windows. If we have a simple entry point, we could let buildout generate a script in the bin folder, as we do for the zopepy script for example.

PR #295, opened 2018-06-11
add configure subtemplate

kakshay21:
see https://github.com/plone/plonecli/issues/22


kakshay21:
Moving this work to plonecli as it makes more sense to keep plonecli stuff at one place

PR #297, opened 2018-06-13
shift changes from buildout to addon

kakshay21:
This PR is shifting those changes which were in buildout subtemplate to addon subtemplate.
see #212

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,10 +4,7 @@ Changelog
 3.3.1 (unreleased)
 ------------------
 
-- Shifting buildout's .gitignore and buildout.cfg template to addon
-  [kakshay21]
-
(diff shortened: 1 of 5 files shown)


CHANGES.rst
```diff
@@ -4,7 +4,12 @@ Changelog
 3.3.1 (unreleased)
 ------------------
 
+- Shifting buildout's .gitignore and buildout.cfg template to addon
```
MrTango: this line is not useful in the changelog, as it is a fix of the change you already mentioned.
Maybe better change the existing entry, to make it clear where this happend:

> Adding plone-compile-resources script in the buildout and extending .gitignore file

kakshay21:
removing the cache worked

Issue #280, opened 2018-04-25
Add sub-template "view" to create views


MrTango:
The sub-template should be able to create a BrowserView or a Dexterity DefaultView.

A view consists of:

- a zcml registration
- a python file per view which contains the view class which as either based on BrowserView or DefaultView class
- a optional template

We could also ask for the output type (HTML or JSON) and generate different python examples.
The json example should render the output in json and also set the correct header to disable Diazo for example.

MrTango:
the structure should be like this:

```
/views/configure.zcml
/views/my_first_view.py
/views/my_first_view.pt
/views/my_second_view.py
/views/my_second_view.pt
```

PR #299, opened 2018-06-18
Add view subtemplate

kakshay21:
see #280

### DIFF
--- a/bobtemplates/plone/view.py
+++ b/bobtemplates/plone/view.py
@@ -1,0 +1,203 @@
+# -*- coding: utf-8 -*-
+"""Generate view."""
+
+from bobtemplates.plone.base import base_prepare_renderer
+from bobtemplates.plone.base import git_commit
+from bobtemplates.plone.base import update_file
+from lxml import etree
+from mrbob.bobexceptions import SkipQuestion
+from mrbob.bobexceptions import ValidationError
+
+import os
+import stringcase
+
+
+def get_view_name_from_python_class(configurator, question):
+    """Generate view default name from python class"""
+    if configurator.variables['view_python_class']:
+        view_class_name = configurator.variables['view_python_class_name']
+        view_generated_name = stringcase.snakecase(view_class_name).replace('_', '-')  # NOQA: E501
+        question.default = view_generated_name
+    else:
+        question.default = 'my-view'
+
+
+def get_template_name_default(configurator, question):
+    if configurator.variables['view_template']:
+        view_url = configurator.variables['view_name']
+        template_default_name = stringcase.snakecase(view_url)
+        question.default = template_default_name
+    else:
+        question.default = 'view'
+
+
+def check_python_class_answer(configurator, question):
+    if not configurator.variables['view_python_class']:
+        raise SkipQuestion(u'No python class, so we skip python class name question.')  # NOQA: E501
+
+
+def check_view_template_answer(configurator, question):
+    if not configurator.variables['view_template'] and not configurator.variables['view_python_class']:  # NOQA: E501
+        raise ValidationError(u'View must at least have a template or a python class')  # NOQA: E501
+    elif not configurator.variables['view_template']:
+        raise SkipQuestion(u'No view template, so we skip view template name question.')  # NOQA: E501
+
+
+def _update_views_configure_zcml(configurator):
+    file_name = u'configure.zcml'
+    directory_path = configurator.variables['package_folder'] + '/views/'
+    file_path = directory_path + file_name
+    configure_example_file_path = configurator.variables['package_folder'] + '/views/configure.zcml.example'  # NOQA: E501
+    file_list = os.listdir(os.path.dirname(directory_path))
+    if file_name not in file_list:
+        os.rename(configure_example_file_path, file_path)
+    namespaces = '{http://namespaces.zope.org/zope}'
+
+    with open(file_path, 'r') as xml_file:
+        parser = etree.XMLParser(remove_blank_text=True)
+        tree = etree.parse(xml_file, parser)
+        tree_root = tree.getroot()
+        view_xpath = "{0}browser:page[@name='{1}']".format(
+            namespaces,
+            configurator.variables['view_name'],
+        )
+        if len(tree_root.findall(view_xpath)):
+            print(
+                '{0} already in configure.zcml, skip adding!'.format(
+                    configurator.variables['view_name'],
+                ),
+            )
+            return
+
+    match_str = '-*- extra stuff goes here -*-'
+
+    if configurator.variables['view_template'] and configurator.variables['view_python_class']:  # NOQA: E501
+        insert_str = """
+  <browser:page
+     name="{0}"
+     for="Products.CMFCore.interfaces.IFolderish"
+     class=".{1}.{2}"
+     template="{3}.pt"
+     permission="zope2.View"
+     />
+        """.format(
+            configurator.variables['view_name'],
+            configurator.variables['view_python_file_name'],
+            configurator.variables['view_python_class_name'],
+            configurator.variables['view_template_name'],
+        )
+
+    if configurator.variables['view_template'] and not configurator.variables['view_python_class']:  # NOQA: E501
+        insert_str = """
+    <browser:page
+       name="{0}"
+       for="Products.CMFCore.interfaces.IFolderish"
+       template="{1}.pt"
+       permission="zope2.View"
+       />
+       """.format(
+            configurator.variables['view_name'],
+            configurator.variables['view_template_name'],
+        )
+
+    if not configurator.variables['view_template'] and configurator.variables['view_python_class']:  # NOQA: E501
+        insert_str = """
+    <browser:page
+       name="{0}"
+       for="Products.CMFCore.interfaces.IFolderish"
+       class=".{1}.{2}"
+       permission="zope2.View"
+       />
+       """.format(
+            configurator.variables['view_name'],
+            configurator.variables['view_python_file_name'],
+            configurator.variables['view_python_class_name'],
+        )
+
+    update_file(configurator, file_path, match_str, insert_str)
+
+
+def _update_configure_zcml(configurator):
+    file_name = u'configure.zcml'
+    file_path = configurator.variables['package_folder'] + '/' + file_name
+    namespaces = '{http://namespaces.zope.org/zope}'
+
+    with open(file_path, 'r') as xml_file:
+        parser = etree.XMLParser(remove_blank_text=True)
+        tree = etree.parse(xml_file, parser)
+        tree_root = tree.getroot()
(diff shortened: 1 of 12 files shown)


bobtemplates/plone/view/views/+view_name_normalized+.pt.example.bob
```diff
@@ -0,0 +1,12 @@
+<html xmlns="http://www.w3.org/1999/xhtml"
+      xmlns:metal="http://xml.zope.org/namespaces/metal"
+      xmlns:tal="http://xml.zope.org/namespaces/tal"
+      xmlns:i18n="http://xml.zope.org/namespaces/i18n"
+      metal:use-macro="context/main_template/macros/master">
+<body>
+    <metal:block fill-slot="content-core">
+        <h1>Sample View</h1>
```
MrTango: add another tag here which actually calls your view method.
Also use i18n-translate here as example.
Remember this code should be also a good basic example of how to write a view.

bobtemplates/plone/view/.mrbob.ini
```diff
@@ -0,0 +1,48 @@
+[questions]
+subtemplate_warning.question = Please commit your changes, before using a sub-template! Continue anyway? [n/y]
+subtemplate_warning.required = True
+subtemplate_warning.default = n
+subtemplate_warning.pre_ask_question = bobtemplates.plone.base:git_clean_state_check
+subtemplate_warning.post_ask_question = mrbob.hooks:validate_choices bobtemplates.plone.base:subtemplate_warning_post_question
+subtemplate_warning.choices = y|n
+subtemplate_warning.choices_delimiter = |
+
+view_python_class.question = Use a python class?
```
MrTango: Better: Should the view have a Python class?

bobtemplates/plone/view/.mrbob.ini
```diff
@@ -0,0 +1,48 @@
...
+view_python_class.help = Do you want python class for this view?
+view_python_class.required = True
+view_python_class.default = y
+view_python_class.post_ask_question = mrbob.hooks:validate_choices mrbob.hooks:to_boolean
+view_python_class.choices = y|n
+view_python_class.choices_delimiter = |
+
+view_python_class_name.question = python class name
```
MrTango: Python not python ;)

bobtemplates/plone/view/.mrbob.ini
```diff
@@ -0,0 +1,48 @@
...
+view_python_class_name.help = Shold be something like 'MyView'
+view_python_class_name.required = False
+view_python_class_name.default = MyView
+view_python_class_name.pre_ask_question = bobtemplates.plone.view:check_python_class
+
+view_name.question = View name
+view_name.help = Should be something like 'My View' (no special characters!)
```
MrTango: The View name is part of the URL and should be something like my-view.

bobtemplates/plone/view/.mrbob.ini
```diff
@@ -0,0 +1,48 @@
...
+view_python_class.choices = y|n
+view_python_class.choices_delimiter = |
+
+view_python_class_name.question = python class name
+view_python_class_name.help = Shold be something like 'MyView'
+view_python_class_name.required = False
+view_python_class_name.default = MyView
+view_python_class_name.pre_ask_question = bobtemplates.plone.view:check_python_class
+
+view_name.question = View name
```
MrTango: Better: View name (part of the URL)

bobtemplates/plone/view/.mrbob.ini
```diff
@@ -0,0 +1,48 @@
...
+view_name.help = Should be something like 'My View' (no special characters!)
+view_name.required = True
+view_name.default = My View
+view_name.pre_ask_question = bobtemplates.plone.base:check_root_folder
+view_name.post_ask_question = bobtemplates.plone.view:check_view_name
+
+view_template.question = Use a page template?
```
MrTango: Should the View have a template file?

bobtemplates/plone/view/.mrbob.ini
```diff
@@ -0,0 +1,48 @@
...
+view_template.help = Do you want a tempalte file for this view?
+view_template.required = True
+view_template.default = y
+view_template.post_ask_question = mrbob.hooks:validate_choices mrbob.hooks:to_boolean
+view_template.choices = y|n
+view_template.choices_delimiter = |
+
+view_template_name.question = page template name
```
MrTango: Better: Template name (without extension)

bobtemplates/plone/view/views/+view_name_normalized+.py.example.bob
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
+
+{{% if view_python_class %}}
+from {{{ package.dottedname }}} import _
+from Products.Five.browser import BrowserView
+
+
+class {{{ view_python_class_name }}}(BrowserView):
+
```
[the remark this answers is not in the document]
kakshay21: Adding call method will force to load that instead of any template if available. Should I put that in the question too?

bobtemplates/plone/view/.mrbob.ini
```diff
@@ -0,0 +1,49 @@
...
+view_template.help = Do you want a template file for this view?
+view_template.required = True
+view_template.default = y
+view_template.post_ask_question = mrbob.hooks:validate_choices mrbob.hooks:to_boolean
+view_template.choices = y|n
+view_template.choices_delimiter = |
+
+view_template_name.question = Template name (without extension)
+view_template_name.help = name of the template file for this view
+view_template_name.default = my_view
```
MrTango: this should be dynamic generated from the Python Class name if we have one ;) otherwise set "view" as the default

MrTango: one thing, use the view_name to generate the default not the class name as we always have the view_name at this point

bobtemplates/plone/view/views/+view_name_normalized+.pt.example.bob
```diff
@@ -0,0 +1,14 @@
+<html xmlns="http://www.w3.org/1999/xhtml"
+      xmlns:metal="http://xml.zope.org/namespaces/metal"
+      xmlns:tal="http://xml.zope.org/namespaces/tal"
+      xmlns:i18n="http://xml.zope.org/namespaces/i18n"
+      metal:use-macro="context/main_template/macros/master">
```
MrTango: define a i18n:domain with the package name

bobtemplates/plone/view.py
```diff
@@ -0,0 +1,233 @@
...
+from bobtemplates.plone.base import update_file
+from lxml import etree
+from mrbob.bobexceptions import SkipQuestion
+from mrbob.bobexceptions import ValidationError
+
+import os
+import stringcase
+
+
+def get_view_name(configurator, question):
```
MrTango: you describe in the doc string way better what the method is doing.
Try to improve the method name, so that it makes more sense.

bobtemplates/plone/view.py
```diff
@@ -0,0 +1,233 @@
...
+    if configurator.variables['view_python_class']:
+        class_name = configurator.variables['view_python_class_name']
+        view_name = stringcase.snakecase(class_name).replace('_', '-')
+        question.default = view_name
+    else:
+        question.default = 'my-view'
+
+
+def get_template_name(configurator, question):
+    if configurator.variables['view_template']:
```
MrTango: this variable name is confusing me. view_name doesn't say anything.
Please go over the names and name them, so that they tell me for what they actually are.

bobtemplates/plone/view.py
```diff
@@ -0,0 +1,233 @@
...
+        view_name = configurator.variables['view_name']
+        template_name = stringcase.snakecase(view_name)
+        question.default = template_name
+    else:
+        question.default = 'view'
+
+
+def check_python_class(configurator, question):
+    if not configurator.variables['view_python_class']:
```
MrTango: same here, dont't know what view_python_class means, is it the name of a class? that would be my frist guess.

bobtemplates/plone/view.py
```diff
@@ -0,0 +1,233 @@
...
+        raise SkipQuestion(u'No python class, so we skip python class name question.')  # NOQA: E501
+
+
+def check_view_template(configurator, question):
```
MrTango: same here, the name could tell me what it does.

kakshay21:
@MrTango I'll rebase it after your review.

bobtemplates/plone/view/views/+view_name_normalized+.py.bob
```diff
@@ -0,0 +1,12 @@
...
+
+{{% if view_python_class %}}
+from {{{ package.dottedname }}} import _
+from Products.Five.browser import BrowserView
+
+
+class {{{ view_python_class_name }}}(BrowserView):
+
+    def __call__(self):
+        return _(u'This view works!')
```
MrTango: The view is not rendering well. I expect the plone UI with your string, not just a string.

bobtemplates/plone/view/.mrbob.ini
```diff
@@ -0,0 +1,49 @@
...
+
+view_template.question = Should the View have a template file?
+view_template.help = Do you want a template file for this view?
+view_template.required = True
+view_template.default = y
+view_template.post_ask_question = mrbob.hooks:validate_choices mrbob.hooks:to_boolean
+view_template.choices = y|n
+view_template.choices_delimiter = |
+
+view_template_name.question = Template name (without extension)
```
MrTango: the template name, does not have any effect, i chose 
--> Template name (without extension) [my_template_only_view]: my-template-only-view

but the template is named: my_template_only_view.pt

Please test this in a test too.
Also add a skeleton test for the view. The the generate view should have at least one test, which is getting the view and calling a method on it. And check the redendered output too.

bobtemplates/plone/view/views/+view_name_normalized+.py.bob
kakshay21: @MrTango For this, we need a template, isn't it?

bobtemplates/plone/view/views/+view_name_normalized+.pt.bob
```diff
@@ -0,0 +1,18 @@
+<html xmlns="http://www.w3.org/1999/xhtml"
+      xmlns:metal="http://xml.zope.org/namespaces/metal"
+      xmlns:tal="http://xml.zope.org/namespaces/tal"
+      xmlns:i18n="http://xml.zope.org/namespaces/i18n"
+      i18n:domain="{{{ package.dottedname }}}"
+      metal:use-macro="context/main_template/macros/master">
+<body>
```
MrTango: check the file name here, it's using the wrong variable, we have the name for the template from a question, we should it ;)

PR #304, opened 2018-06-23
Add update_locale script

kakshay21:
see #45


bobtemplates/plone/addon/src/+package.namespace+/+package.name+/locales/update.py.bob
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
+import subprocess
+
+
+domain = '{{{ package.dottedname }}}'
+i18ndude = '../../../../bin/i18ndude'
```
MrTango: this path should be dynamic.
You know where the package root is and from there it should be ./bin/i18ndude.
But check also the context of the script when you call it via ./bin/update_locales

PR #319, opened 2018-08-30
Fix theme_package fonts

erral:
The Roboto font files included in the `theme_package` template don't match the CSS registrations from barceloneta (also included in the template).

The font files are capitalized and the CSS registrations are done using lowecased font names.

With this PR the fonts are now lowercased like they are in the original plonetheme.barceloneta

See fonts here: https://github.com/plone/bobtemplates.plone/tree/master/bobtemplates/plone/theme_package/src/%2Bpackage.namespace%2B/%2Bpackage.name%2B/theme/less/roboto

And CSS registration here:

https://github.com/plone/bobtemplates.plone/blob/master/bobtemplates/plone/theme_package/src/%2Bpackage.namespace%2B/%2Bpackage.name%2B/theme/barceloneta/less/fonts.plone.less


jensens:
Wow, we include whole barceloneta here? Anyway, the change itself is fine.

MrTango:
@jensens yes that was the way i choose back then. There are many ways to get customize baceloneta, but most of them work either TTW or with the Grunt setup or only when you are online. This could be improved with a update barceloneta resources, so that we have it always up to date. But i don't know if for people using this it's so importent that it works with the Grunt setup offline. Maybe it's enought to use it TTW. In this case the whole setup is much simpler. @erral wanted to work on a barceloneta based theme subtemplate, which works like the current basic "theme"-subtemplate but is meant to customize barceloneta like this template does. 

Would be nice to have this and I'm ok with TTW only or online Grunt setup which uses github urls to include barceloneta resources. Also @tmassman worked a lot on the theming story, maybe he can help or give advice here.

erral:
@MrTango I have the barceloneta subtemplate (for addon template) almost ready, I need to do a couple of checks and will add a PR.

PR #321, opened 2018-09-28
Commented out rule in backend.xml that caused the scripts being pulle…

### LINKED ISSUE #320: backend.xml pulls the all scripts again, causing non deterministic JS errors
This rule:
`    <after css:theme-children="head" css:content="head script" />`
is present in both rules.xml and in backend.xml, so that causes to be added twice:

screenshot 2018-09-28 at 10 28 52

In case that you are including backend.xml in your rule set.

This should be changed in barceloneta (https://github.com/plone/plonetheme.barceloneta/issues/158) and in the templates as well.

sneridagh:
…d twice in case that you are using it to style backend views.

This fixes #320

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,13 @@ Changelog
 - Use --ignore-submodules in git_clean_state_check
   [MrTango]
 
+- Commented out rule in backend.xml that caused the scripts being pulled twice,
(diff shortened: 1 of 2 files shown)


sneridagh:
Cool! Thanks!

PR #323, opened 2018-10-16
Add a theme_barceloneta subtemplate for addon template

### LINKED ISSUE #240: Full-featured theme subtemplate
We have been discussing about the `theme_package` and the `theme` subtemplate at work, and think that could be nice to have a subtemplate that adds a theme but not an empty one but a similar one to the one created by the `theme_package` template.

This way you can add a complete theme into an addon.

This has come in a conversation that we had where we try to create just one package for our client websites, because almost always the content types created for a customer are tightly coupled to the theme and viceversa.

If this is interesting we can work on this

@nruiz @libargutxi

erral:
This PR adds a new subtemplate called `theme_barceloneta` to the existing `addon` template. 

This way one can add a barceloneta-based theme to an existing addon.

Fixes #240


bobtemplates/plone/theme_barceloneta/DEVELOP_THEME.rst.bob
```diff
@@ -0,0 +1,42 @@
...
+
+    $ grunt plone-bsync
+
+It will run a Grunt browser sync task in a new browser window and will reload after every change of the less files in the less folder. If your Plone is not running on port 8080 you have to adjust the proxy option in the Gruntfile.js.
+
+If you prefer to do a one time compile of the less files you can run::
+
+    $ grunt compile
+
+or use the helper script from buildout::
```
MrTango: Please check this, i don't think that it's working so far, because you don't update the existing buildout config, which is coming from the addon template.

bobtemplates/plone/theme_barceloneta/DEVELOP_THEME.rst.bob
```diff
@@ -0,0 +1,42 @@
...
+
+    $ ./bin/grunt-task-compile
+
+Actually on a buildout run buildout will run the npm-install and the grunt-task-compile script for you. So you only need the commands above while developing gin the filesystem.
```
MrTango: same here, alternatively we could skip it but document how it can be archived manually or leave it out at all.

erral: You are right.

I have removed the references to the buildout-way and explained how to do it using regular npm install.

Thanks,

Issue #335, opened 2018-10-23
bobtemplate.plone breaks search for special characters in Plone


pbauer:
See https://github.com/plone/Products.CMFPlone/issues/2599

MrTango:
I had to revert my fix, it's used by the portlet template. In my opinien this has to be fixed in Plone, so that Unidecode>=0.04.16 is no longer a problem, as this is an already many years old version, so it's time to move on. Even the odest version of slugify requires a newer version.

I'm open to any replacement for the library.

pbauer:
Fixed with plone/bobtemplates.plone@1a76769 in `bobtemplates.plone = 3.5.2`

PR #339, opened 2018-10-29
Remove dependency on python-slugify

pbauer:
Fixes https://github.com/plone/Products.CMFPlone/issues/2599 and https://github.com/plone/bobtemplates.plone/issues/335

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,8 +8,7 @@ Changelog
   [MrTango]
 
 - Remove unnecessary dependency to python-slugify, fixes: #335 and #337
-  [MrTango]
-- Nothing changed yet.
+  [pbauer]
 
 - Add docs for view, viewlet and portlet
   [kakshay21]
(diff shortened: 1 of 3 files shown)


pbauer:
@MrTango I agree that it needs to be fixed in Plone and that is is probably simple. But if we only do that then only the newest Plone (once it will be released) can use bobtemplates.plone and that would be monumentally stupid. Then also transform the string to ascii there are many options, I don't really care. But it can't stay like this.

MrTango:
Why is it so important to have bobtemplates.plone in the development.cfg or in buildout in general?
I get that it is currently comming like this in the unifiedinstaller, but do we have a problem in saying remove it there for now and you don't have a problem? After removing it from there and runnign buildout again everything shoudl work, right? I'll have a look at the implementation in bobtemplates.plone as soon as I can make some time for that. Maybe we should just restrict the portlet name for now.

MrTango:
This is already released and solved, so i close this PR

Issue #338, opened 2018-10-29
bobtemplates.plone has no buildout.cfg


pbauer:
In cb1d3bd1f9be02cebd2f04215734aea2908ddc35 the `buildout.cfg` was removed my @loechel. Why? How would I start developing it? https://github.com/plone/bobtemplates.plone/blob/master/DEVELOP.rst still references a buildout.cfg, so the docs are broken.

MrTango:
The docs are here: https://github.com/plone/bobtemplates.plone/blob/master/docs/develop.rst the DEVELOP.rst needs to be updated or deleted like @loechel started already.

PR #340, opened 2018-10-29
cleanup root docs

loechel:
remove Develop.rst and add a paragraph in README.rst

solve #338


loechel:
@gforcada could you check, why @mister-roboto yells at me?

MrTango:
@loechel can you please add a changelong entry ;)

MrTango:
I think this changes are obsolete with my last changes.

PR #341, opened 2018-10-30
add a template for creating a migration package

cdw9:
I'm building a Transmogrifier training for training.plone.org that goes along with this template.


loechel:
@MrTango can you have a look, what to do with this general Python regression? 

Currently I am actually willing to accept this merge request with failing tests, as they should appear also on master and all other pull requests.

cdw9:
Discussed this template with @MrTango - instead of making the migration template a part of bobtemplates.plone, I will create a separate repository/package for it. This will also allow us to make multiple migration templates, that can have different default pipelines depending on your exported data.

PR #354, opened 2019-02-21
Use implementer decorator for python3 compatibility

erral:
Using ```implements``` is not allowed in python 3 and Plone 5.2:

```
    TypeError: Class advice impossible in Python3.  Use the @implementer class decorator instead.
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,7 +9,8 @@ Changelog
 3.6.0b2 (unreleased)
 --------------------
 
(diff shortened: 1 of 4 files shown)


erral:
I fixed the test.

You may want to look at the changelog, there is something there coming from a previous commit...

We may want to migrato to towncrier: https://community.plone.org/t/using-towncrier-for-changes-rst-in-plone-packages/6486

PR #356, opened 2019-02-27
Fix behavior subtemplate: use separate marker interface

fredvd:
Register marker in the behavior zcml and adapt content to the marker.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,10 @@ Changelog
 3.6.1 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Fix behavior template: use separate marker interface, register marker in the
+  behavior zcml and adapt content to the marker, not to IDexterityContent. For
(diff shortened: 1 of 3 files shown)


fredvd:
Checked locally with a dummy package and default behavior:

```
    <plone:behavior
        name="collective.foo.attachment_type"
        title="AttachmentType"
        description="This behavior provides..."
        provides=".attachment_type.IAttachmentType"
        factory=".attachment_type.AttachmentType"
        marker=".attachment_type.IAttachmentTypeMarker"
        />
```

```
class IAttachmentTypeMarker:
    pass

@provider(IFormFieldProvider)
class IAttachmentType(model.Schema):
    """
    """

    project = schema.TextLine(
        title=_(u'Project'),
        description=_(u'Give in a project name'),
        required=False,
    )


@implementer(IAttachmentType)
@adapter(IAttachmentTypeMarker)
class AttachmentType(object):
    def __init__(self, context):
        self.context = context
```

Issue #345, opened 2018-11-21
Use behavior shortnames 


jensens:
Use shortnames like ``plone.locking`` instead of the dotted name.
https://github.com/plone/bobtemplates.plone/blob/21fb444d36fc79890e7abfd3d9f403c9720a0aa2/bobtemplates/plone/content_type/profiles/default/types/%2Bdexterity_type_fti_file_name%2B.xml.bob#L57-L94

jensens:
see https://docs.plone.org/external/plone.app.dexterity/docs/reference/standard-behaviours.html

PR #359, opened 2019-03-03
Fix behaviors fti

jensens:
- Use behavior shortnames in FTI. see #345 

- Do not add dexterity as dependency in Plone 5.x due to the fact it is already in Plone core.

- Do not try to use the main supermodel schema class as behavior.
  This can not work and will lead to warnings.

### DIFF
--- a/bobtemplates/plone/content_type/profiles/default/types/+dexterity_type_fti_file_name+.xml.bob
+++ b/bobtemplates/plone/content_type/profiles/default/types/+dexterity_type_fti_file_name+.xml.bob
@@ -51,9 +51,6 @@
 
   <!-- Enabled behaviors -->
   <property name="behaviors" purge="false">
-{{% if not dexterity_type_create_class %}}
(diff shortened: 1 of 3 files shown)


bobtemplates/plone/content_type/profiles/default/types/+dexterity_type_fti_file_name+.xml.bob
```diff
@@ -51,21 +51,18 @@
 
   <!-- Enabled behaviors -->
   <property name="behaviors" purge="false">
-{{% if not dexterity_type_create_class %}}
-    <element value="{{{ package.dottedname }}}.content.{{{ dexterity_type_name_normalized }}}.I{{{ dexterity_type_name_klass }}}"/>
-{{%  endif %}}
```
jensens: @MrTango This does not make any sense to me, so I remove it here. The schema is never registered as a behavior nor does it make sense to do so. 
If you or anyone else have any idea why someone tried to do this, please explain.

jensens: btw, the plain-schema-as-a-behavior is deprecated in plone.dexterity and marked for removal with the next major of plone.dexterity

PR #373, opened 2019-04-30
5.2 cleanup theme_barceloneta WIP

agitator:


bobtemplates/plone/theme_barceloneta/theme/barceloneta/BARCELONETA-VERSION.rst
```diff
@@ -0,0 +1 @@
+2.1.3
```
MrTango: i would move this file one level up

djowett:
Does this apply to 5.1 as well as 5.2?

agitator:
@djowett barceloneta 2.1.2 should work with Plone 5.1 and 5.2, so these changes should also work for both version.
Not sure though why http://dist.plone.org/release/5.1.6-pending/versions.cfg wasn't updated with  https://github.com/plone/buildout.coredev/blob/5.1/versions.cfg#L290 yet

djowett:
Thanks @agitator

balavec:
There are also some file permission changes from 100644 → 100755. Is this intentional?

agitator:
Todo:
- fix robot font path
- update Readme

PR #393, opened 2019-06-20
Black config added

loechel:


tox.ini
```diff
@@ -98,6 +99,24 @@ commands =
...
+    black --line-length=79 --check --diff -v --skip-string-normalization bobtemplates setup.py
+
+[testenv:black-enforce]
+basepython = python3.7
+skip_install = True
+deps =
+    black
+
+commands =
+    black -v --line-length=79 --skip-string-normalization bobtemplates setup.py
```
ale-rt: Just in case... When discussing the black PLIP (https://github.com/plone/Products.CMFPlone/issues/2754) it was decided to use black with all the defaults.

loechel: I wanted that for the first run, as it confirms with flake8 settings. Will do more basic setting later to see different results.

loechel: I don't want to have to many changes on one commit, I do it step by step to have black defaults. Before we did have already used black for some checks, but with those specific config.

MrTango:
@loechel still want to merge this? is it ready or do you want to tune it?

loechel:
Sorry I have been too busy.

Maybe keep that merge Request open, and make two new out of it, one the real config changes and one that applied black.

MrTango:
Since i already integrated black support meanwhile in version 6.x i'll close this, but keep the branch in case you need anything from it. please remove it, if you don't need it anymore.

PR #398, opened 2019-07-22
Change fonts mixin to bob template and use theme normalized_name

balavec:
Change fonts mixin to bob template and use theme normalized_name to fix the path

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changelog
 4.1.1 (unreleased)
 ------------------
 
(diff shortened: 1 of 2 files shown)


balavec:
@agitator who does usually merge this to master? Do we need to update the version or something?

agitator:
If the pr is approved, you can merge it yourself ;-)
With Plone we do https://semver.org 
For the theme_barceloneta I'd see the changes as a feature, but @MrTango should decide in this cas

PR #406, opened 2019-09-09
Fix travis setup: geckodriver & coveralls

erral:
As reported on Community Forum, when a newly created package is run in Travis a 'geckodriver is missing in PATH' error is created. 

https://community.plone.org/t/travis-ci-test-failed-with-geckodriver-executable-needs-to-be-in-path/9071

Also, run coveralls from the created bin directory.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changelog
 4.1.4 (unreleased)
 ------------------
 
+- Run coveralls in the correct path
+  [erral]
+
 - Fix 'geckodriver' error when running tests in Travis.
   [erral]
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,8 @@ Changelog
 4.1.4 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Fix 'geckodriver' error when running tests in Travis.
+  [erral]
 
 
 4.1.3 (2019-08-31)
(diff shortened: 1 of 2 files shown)


erral:
I just saw the comments on community.plone.org and I was also hit by the same error some days ago. So I just copied over the solution provided there, and as you say I don't know what happens here because this used to work.

jensens:
TL;DR: For Python 3.7 we can get it via apt. For older Pythons we need to fetch it via wget.

See my new comment at 
https://community.plone.org/t/travis-ci-test-failed-with-geckodriver-executable-needs-to-be-in-path/9071/5?u=jensens

MrTango:
this way seems to be used in other packages too: https://github.com/plone/plone.app.mosaic/blob/master/.travis.yml#L30
But you are right, we should find a better method.
I'll merge the fix for now, but feel free to improve it afterwards.

jensens:
I just read, we could use xenial by default, because it support all Python versions needed by us, see https://docs.travis-ci.com/user/reference/xenial/#python-support. Worth I try?

jensens:
Aaaaand: Instead of the xenial, nowadays bionic is supported with all our Python version needed. 
https://docs.travis-ci.com/user/reference/bionic/#python-support

MrTango:
I just updated the master with dist: bionic and installing geckodriver from apt ;).
Fell free to test it out.

Issue #408, opened 2019-09-16
Travis cache settings seem wrong


mauritsvanrees:
Currently [.travis.yml.bob](https://github.com/plone/bobtemplates.plone/blob/master/bobtemplates/plone/addon/.travis.yml.bob) has:

```
cache:
  pip: true
  directories:
  - eggs
  - downloads
...
before_install:
...
  - echo "download-cache = $HOME/buildout-cache/downloads" >> $HOME/.buildout/default.cfg
  - echo "eggs-directory = $HOME/buildout-cache/eggs" >> $HOME/.buildout/default.cfg
```

So shouldn't the `cache: directories` setting contain `buildout-cache` instead of `eggs` and `downloads`?
It is a bit hard to say for sure, because you don't see what ends up in the Travis cache. I thought I had seen a link where you could download or inspect a cache file, but I do not see it now.

mauritsvanrees:
TL;DR: we should simply cache the `eggs` directory, and not set a buildout `download-cache` or `eggs-directory`. See bottom.

I have experimented a bit with the cache setup in [collective.denyroles](https://github.com/collective/collective.denyroles). Have a look at the buildout time in [Travis](https://travis-ci.org/collective/collective.denyroles/builds).

The first build, with only one job in the matrix, took 6 minutes, and the next one less than 2 minutes, so the cache was working at that point. [.travis.yml](https://github.com/collective/collective.denyroles/blob/e64002298611f14ea4341a3a0a231e348293a721/.travis.yml) was:

```
cache:
  pip: true
  directories:
  - eggs
  - downloads
...
bin/buildout -N buildout:download-cache=downloads
```

Slightly different than current bobtemplates.plone, but working.

Later I added more jobs to the matrix, so the builds obviously started taking longer, but I also switched to what bobtemplates.plone currently has, or variants thereof. See [this commit](https://github.com/collective/collective.denyroles/commit/c8b82aeb58b82045744bc9ac774930a46fd0f748). That build took 12 minutes, probably with the previous cache still somewhat working, and later builds 17 to 18 minutes. [.travis.yml](https://github.com/collective/collective.denyroles/blob/c8b82aeb58b82045744bc9ac774930a46fd0f748/.travis.yml):

```
cache:
  pip: true
  directories:
  - buildout-cache
...
before_install:
...
  - echo "[buildout]" > $HOME/.buildout/default.cfg
[...]

PR #409, opened 2019-09-18
Fixed Travis caching.

mauritsvanrees:
Downloads do not need to be cached.
And for the eggs we were caching the wrong directory.
Fixes issue #408.

Also let plone/theme_package use a bionic machine on Travis, so the config is more similar to plone/addon.


MrTango:
maybe let's see how this has to work when we have made the switch to tox in travis setup. tox is already there we but don't run it on travis so far.

PR #412, opened 2019-10-21
virtualenv generate .Python file on OS X

fredvd:
put in gitignore


MrTango:
the test are failing, maybe you based on a broken commit. please rebase on master and see if that fixes you PR.

balavec:
That is added in the #416.

PR #416, opened 2019-11-01
Make code-analysis actually run on code, fix flake8 issues

balavec:


skeleton-tests/test_addon_theme.py
```diff
@@ -30,7 +30,7 @@ def test_addon_theme(tmpdir, capsys, config):
 
     # generate template addon:
     config.template = 'addon'
-    config.package_name = 'plonetheme.task'
```
MrTango: I think we should stay with plonetheme.* here, because thats what most people we use.
I there are any isort issue because of the the name, i would prefer to resort them for the theme templates, if doable.

balavec: @MrTango I will check.

MrTango: ?

balavec: @MrTango I was a bit caught up. Let me dive into it.

balavec: I updated the name. Let me know how that works for you.

balavec:
@MrTango your comment about tox doing the linting - did you have in mind something like pylint?

balavec:
@MrTango I updated the PR and now also linting is setup. I also had to update the buildout.cfg.bob file (former buildout.cfg) to install the correct version of Plone. Can you please review it?

bobtemplates/plone/addon/test_plone52.cfg
```diff
@@ -1,7 +1,7 @@
 [buildout]
 
 extends =
-    https://raw.github.com/collective/buildout.plonetest/master/test-5.2.x.cfg
+    https://raw.githubusercontent.com/collective/buildout.plonetest/master/test-5.2.x.cfg
```
thet: :+1:

bobtemplates/plone/addon/test_plone51.cfg
```diff
@@ -1,7 +1,7 @@
 [buildout]
 
 extends =
-    https://raw.github.com/collective/buildout.plonetest/master/test-5.1.x.cfg
+    https://raw.githubusercontent.com/collective/buildout.plonetest/master/test-5.1.x.cfg
```
thet: :+1:

bobtemplates/plone/addon/test_plone50.cfg
```diff
@@ -1,7 +1,7 @@
 [buildout]
 
 extends =
-    https://raw.github.com/collective/buildout.plonetest/master/test-5.0.x.cfg
+    https://raw.githubusercontent.com/collective/buildout.plonetest/master/test-5.0.x.cfg
```
thet: :+1:

CHANGES.rst
```diff
@@ -4,12 +4,16 @@ Changelog
...
 
-- Nothing changed yet.
+- Make tox run lint, refactor plone versions, fix import sorting for plonetheme.task package.
+  [balavec]
 
 
 5.0.4 (2019-11-28)
 ------------------
 
+- Make code-analysis actually run on code, fix flake8 issues
```
thet: That should go probably up to 5.0.5

bobtemplates/plone/addon/tox.ini
```diff
@@ -22,7 +22,7 @@ commands =
     {envbindir}/buildout -c {toxinidir}/{env:version_file} buildout:directory={envdir} buildout:develop={toxinidir} install test robot code-analysis
     coverage run {envbindir}/test -v1 --auto-color {posargs}
     # coverage run {envbindir}/test -v --all -t robot {posargs}
-    {envbindir}/code-analysis
+    {envbindir}/code-analysis {toxinidir}/src
```
thet: :+1:

bobtemplates/plone/theme_package/.mrbob.ini
```diff
@@ -37,7 +37,7 @@ package.git.init.choices_delimiter = |
 
 plone.version.question = Plone version
 plone.version.help = Which Plone version would you like to use?
-plone.version.default = 5.1
+plone.version.default = 5.2
```
thet: :+1:

bobtemplates/plone/theme_package/buildout.cfg.bob
```diff
@@ -1,5 +1,5 @@
 [buildout]
-extends = http://dist.plone.org/release/{{{ plone.version }}}/versions.cfg
+extends = http://dist.plone.org/release/{{{ plone.version }}}-latest/versions.cfg
```
thet: I'm :+1: with this change. Others might want to pick it explicitly to a specific version, but IMO that would get out of date quickly.

tox.ini
```diff
@@ -63,10 +63,10 @@ commands =
...
 setenv =
     COVERAGE_FILE=.coverage.{envname}
-    Plone43: VERSION=4.3-latest
-    Plone50: VERSION=5.0-latest
-    Plone51: VERSION=5.1-latest
-    Plone52: VERSION=5.2-latest
+    Plone43: VERSION=4.3
+    Plone50: VERSION=5.0
+    Plone51: VERSION=5.1
+    Plone52: VERSION=5.2
```
thet: ... not sure about these changes.

CHANGES.rst
MrTango: no thats release like this was a more general fix, the new fixes make sure that we use the generated tox config and let them run on generated packages, to find problems in the generated packages earlier.

skeleton-tests/conftest.py
```diff
@@ -19,8 +19,8 @@ def config():
     skeleton_tox_env_parts = os.environ.get('ENVNAME').split('-')
     config = Config()
     config.verbose = bool(os.environ.get('VERBOSE'))
-    config.version = os.environ.get('VERSION', '5.1-latest')
-    config.skeleton_tox_env = '{0}-{1}'.format(
+    config.version = os.environ.get('VERSION', '5.2')
```
MrTango: why not 5.2-latest?

skeleton-tests/test_addon_theme.py
```diff
@@ -47,6 +47,15 @@ def test_addon_theme(tmpdir, capsys, config):
         os.path.join(tmpdir.strpath, config.package_name),
     )
 
+    subprocess.call(
```
MrTango: It would be better, if we integrate the isort call into the bobtemplate hook so that the test will also be ok after the user generates the package. This fix here only makes our test not fail.

MrTango: could you integrate it into base.py and call it in the post_render https://github.com/plone/bobtemplates.plone/blob/5fe1c59211633078dc5e795adad7781349e7e4be/bobtemplates/plone/addon.py#L138
We could also run this in other sub-templates. Because we never know what the packages names are and need to fix the isort after creating code in general. So all sub-templates need to call this in the post renderer hook.

tox.ini
MrTango: smae here. @balavec do you have a reason to change this? If not i would keep it like it was.

bobtemplates/plone/addon/buildout.cfg.bob
```diff
@@ -3,7 +3,4 @@
 # use this extend one of the buildout configuration:
 extends =
 # -*- mrbob: extra extends -*-
-#    test_plone43.cfg
-#    test_plone50.cfg
-#    test_plone51.cfg
-    test_plone52.cfg
+    test_plone{{{ plone.version_digits }}}.cfg
```
MrTango, line 6: i would keep it like it was, this is not changing too much, only bigger version numbers and some may want to use switch to different versions too.

bobtemplates/plone/addon/tox.ini
MrTango: i removed this reacently completely, we need only tox for linting and testing. using both tools is confusing.

MrTango:
@balavec i did most of these changes now in a diffenrent branch together with other improvements of the package tox setup.
I would say we can close these PR, feel free to pick some changes if they are still missing in master.

PR #426, opened 2020-03-16
fix indexer template

### LINKED ISSUE #425: ZCML created by indexer subtemplate is not correct
The ending `</configure` is not closed correctly and the main configure loads a package instead the file

erral:
This fixes #425

### DIFF
--- a/bobtemplates/plone/indexer.py
+++ b/bobtemplates/plone/indexer.py
@@ -37,7 +37,7 @@ def _update_indexers_configure_zcml(configurator):
     match_xpath = "zope:include[@package='.{0}']".format(configurator.variables['indexer_name'])
     match_str = '-*- extra stuff goes here -*-'
     insert_str = """
-  <include package=".{0}" />
+  <include file="{0}.zcml" />
 """.format(configurator.variables['indexer_name'])
     update_configure_zcml(
         configurator,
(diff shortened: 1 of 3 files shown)


bobtemplates/plone/indexer.py
```diff
@@ -37,7 +37,7 @@ def _update_indexers_configure_zcml(configurator):
     match_xpath = "zope:include[@package='.{0}']".format(configurator.variables['indexer_name'])
     match_str = '-*- extra stuff goes here -*-'
     insert_str = """
-  <include package=".{0}" />
+  <include file="{0}.zcml" />
 """.format(configurator.variables['indexer_name'])
```
ale-rt: Shouldn't this just be `<include package=".indexers" />`?

erral: No, what you say is done some lines before, here: https://github.com/plone/bobtemplates.plone/blob/0e925fa682ae59decd5807e3250fb387713ff248/bobtemplates/plone/indexer.py#L19

ale-rt: Ok, nevermind then :)
I just have no clue of what is happening here!
Please ignore me :)

erral: This exact method modifies the `configure.zcml` file created inside the `indexers` folder and adds a reference to the newly created indexer. But it adds as a "include package" instead of "include file", that's all.

PR #450, opened 2020-10-29
switch from registry.xml to registry folder

MrTango:


tisto:
@MrTango not sure if this is what I think it is. Though, one reason to move to p.a.registry control panels and settings was that spreading settings across multiple files became impossible to manage and we thought that a single file, no matter how big it might become, is the better option. Anyways, I do not use bobtemplates.plone, so this is just my personal recommendation I guess and you are free to do whatever you think is best.

mauritsvanrees (approved):
For me, it is fine to have multiple files in a registry folder. I suspect it also makes things easier for mr.bob: when we add something that requires a change in the registry, we can simply create a new file, instead of editing an existing file.

I am not sure if there is currently code in bobtemplates.plone that adds to the registry. That might need a change then. But the tests pass, so should be fine.

jensens:
+1000 - this is what I change first after creating a project. @tisto I do not want to edit a 1000+ lines XML file, really. And this happens fast in a project.

jensens (approved):

MrTango:
thx for your feedback.
In general you always can use both wys in parallel.
You can still  have the registry.xml and the folder.
For bobtemplates.plone the folder makes it way easier to create new registrations. 
And i prefer simple files over a long single file.

PR #463, opened 2021-05-11
do not create an empty docstring for a content-type class

erral:
I have been bitten by this several times these weeks: when adding an empty docstring to a content-type class, Zope refuses to render the view for it, even the edit form for it, just because the docstring is empty.

If you are using some linting configuration that just converts the empty multiline comment to a one-line empty comment you do not have the docstring and the content items can't be rendered and neither accessed through the ZMI, weird.

This PR adds a dummy docstring to the content-type to avoid this strange errors.


mauritsvanrees (approved):
LGTM.

I have been bitten by this too. Well, I think the content type was not created by bobtemplates, but manually, or otherwise years ago. Recently I ran `black`, this removed the whitespace in the otherwise empty docstring, and it content type could not be viewed anymore.

erral:
Yes, 'black' is to blame here

PR #468, opened 2021-06-08
use package's i18n_domain to translate behaviors

frisi:


frisi:
sorry for the noise. i18ndude happily extracted the msgids but i just recovnized that `dexterity-types/<type>/@@behaviors` does pick translations from the plone domain.

PR #471, opened 2021-09-02
Black/flake8/pylint compatibility

erral:
Add black and flake compatibility files to be consistent with the checks made when using black


erral:
Sorry for the noise, this was not meant to be a PR for bobtemplates.plone

PR #472, opened 2021-10-05
add a controlpanel subtemplate

erral:
Tests are missing, but the subtemplate works great!


MrTango:
Nice, would you mind fixing the tests for Python 3.7? I think it's just linting.
Then we can merge it.
thx a lot

PR #504, opened 2022-05-04
Remove install of unnecessary system libraries on Github Actions

wesleybl:


mauritsvanrees (approved):
I think this was needed in some cases in Plone 4.3 and 5.0 for building lxml. Should be fine to remove yes.

PR #508, opened 2022-07-13
Fix import of ComponentLookupError in tests

MrTango:


mauritsvanrees (approved):
LGTM.
But this would fail on Plone 4.3. Do we still support that here? I have not kept track.

MrTango:
@mauritsvanrees not in main, main and the 6.x versions are Py3 only.

PR #512, opened 2022-08-03
run tests on windows and osx too

jensens:


jensens:
Yay, would be nice to get this one working. Thanks for taking over @mamico

jensens:
outdated

Issue #520, opened 2022-10-15
improve namespace support


MrTango:
make sure that we can create packages in different formats:
- collective.mypackage
- collective.plone.myaddon
- mypackage
- my-package
- collective.my-package

mamico:
Manually tested with 

```
% plonecli -V
        plonecli : 2.4
        bobtemplates.plone: 6.0b16.dev0
 % plonecli create addon NAME
 % cd NAME
 % plonecli build test
```

* collective.mypackage - works
* collective.plone.myaddon - works

* `mypackage` and  `my-package` fail during `plonecli create`

```
% plonecli create addon mypackage
...
Error: 'mypackage' is not a valid packagename.
Please use a valid name (like collective.myaddon or plone.app.myaddon)
```

* `collective.my-package` fails during `plonecli build`

```
error in collective.my-package setup command: 'NoneType' object has no attribute 'group'
While:
  Installing.
  Processing develop directory '.../collective.my-package/.'.

An internal error occurred due to a bug in either zc.buildout or in a
recipe being used:
Traceback (most recent call last):
  File ".../collective.my-package/venv/lib/python3.8/site-packages/zc/buildout/buildout.py", line 2250, in main
    getattr(buildout, command)(args)
  File ".../collective.my-package/venv/lib/python3.8/site-packages/zc/buildout/buildout.py", line 717, in install
    installed_develop_eggs = self._develop()
  File ".../collective.my-package/venv/lib/python3.8/site-packages/zc/buildout/buildout.py", line 960, in _develop
    zc.buildout.easy_install.develop(setup, dest)
  File ".../temp/collective.my-package/venv/lib/python3.8/site-packages/zc/buildout/easy_install.py", line 1120, in develop
    call_subprocess(args)
  File ".../collective.my-package/venv/lib/python3.8/site-packages/zc/buildout/easy_install.py", line 172, in call_subprocess
    raise Exception(
Exception: Failed to run command:
'.../collective.my-package/venv/bin/python3', '/tmp/tmp73x59m63', '-q', 'develop', '-mN', '-d', '.../collective.my-package/develop-eggs/tmpgn8z91m3build'
```

mamico:
closed by #522

PR #522, opened 2022-10-20
improve namespace support #520

mamico:
see https://github.com/plone/bobtemplates.plone/issues/520


MrTango (approved):
Looks good, great work @mamico, thx

PR #525, opened 2022-10-25
improve form template

MrTango:
- use cmf.ManagePortal permission by default in forms
- run applyChanges in forms by default


mamico (approved):
LGTM

PR #526, opened 2022-10-30
add a starter template

erral:
With this starter template we can initialize the Plone site with some sane defaults. This sub-template will be used by cookiecutter-plone-starter

Right now it only sets the site title and configures the default languages.

See https://github.com/collective/cookiecutter-plone-starter/pull/25 for reference.


erral:
@ericof @jensens if you think that more things should be added to the default template when creating a Plone site with cookiecutter-plone-starter, we will add them here.

jensens:
I have no strong opinion here. Less is more IMO, so this looks fine.

MrTango:
If we have this in plonecli/bobtemplates.plone we should add docs and describe better what it does.
The name starter is rather generic.
It might be more clear when you use it from the starter template, but it could also be used inside plonecli directly.
Another option would be, that we deliver this sub-template via a new bobtemplates package which also does the registration for plonecli. That would be the better option, if this template is only useful for the starter-template.
Not sure if thats the case.

erral:
I have renamed the template to be `plone_site_initialization` I think the name is better that the starter one. We may add in the future more configuration items here.

The aim of this subtemplate is to be able to initialize a Plone site using an addon, this way, when installed, the addon can set the site title, available langues, or whatever. It's a kind of "policy product configuration", but I have never liked the "policy" name for this :/

MrTango:
@erral i mannaged to finally fix the build. 
But i have a wish regarding the naming.
I like the plone_site_initialization, it says what it does.
But so far this is only the alias in plonecli.
We should rename the whole template as follows:

folder: /bobtemplate/site_initialization
global name in setup: plone_site_initialization
alias in plonecli: site_initialization

this way it's all clean from the start ;)

erral:
I have renamed everything to site_initialization :)

setup.py
```diff
@@ -72,22 +72,28 @@
             "plone_addon = bobtemplates.plone.bobregistry:plone_addon",
             "plone_behavior = bobtemplates.plone.bobregistry:plone_behavior",
             "plone_buildout = bobtemplates.plone.bobregistry:plone_buildout",
-            "plone_content_type = bobtemplates.plone.bobregistry:plone_content_type",
+            "plone_content_type =" " bobtemplates.plone.bobregistry:plone_content_type",
```
MrTango, line 75: was this intended?

MrTango (changes requested):
looks good, but the changes in setup.py look ood, please explain why we should write the string like this.

setup.py
MrTango: ?

erral: it looks like some lines were converted to 2 lines and then when converting back to one line, the quotes weren't removed.

I have removed them.

erral:
The strange string were removed.

bobtemplates/plone/site_initialization/.mrbob.ini
```diff
@@ -0,0 +1,24 @@
+[questions]
+; subtemplate_warning.question = Please commit your changes, before using a sub-template! Continue anyway? [N/y]
```
MrTango, line 2: do we need these line, or can we clean it up?

MrTango (approved):

bobtemplates/plone/site_initialization/.mrbob.ini
MrTango: ok, if it's a sub template, it should be there, otherwise you risk overwriting files and have no way going back.

erral: Yes, I finally included them

erral:
I will try to take care of this during the Conference sprint

erral:
@MrTango I finally removed the additions I made to include Python 3.11 in the test setup and the PR is green.

Including Python 3.11 should be a PR on itself if wanted.

MrTango (approved):

Issue #531, opened 2022-11-24
Subtemplates are not working properly on windows


me-kell:
Subtemplates are not working properly on windows due to a wrong path calculation.

The function [`bobtemplates.plone.base. base_prepare_renderer`](https://github.com/plone/bobtemplates.plone/blob/9bf1efef7c735ecec99d6e195aa5b10fa0e3dbfa/bobtemplates/plone/base.py#L437) uses `/` to split paths. Paths on windows use `\` and are not splitted. Resulting in wrong paths and failing to generate the templates.

https://github.com/plone/bobtemplates.plone/blob/9bf1efef7c735ecec99d6e195aa5b10fa0e3dbfa/bobtemplates/plone/base.py#L437

The solution is to replace `/` by `os.path.sep`

This has already been reported on https://github.com/plone/bobtemplates.plone/issues/510#issue-1315957061.

I can confirm this solves the problem in Plone-5.2.x and Plone-6.0.0rc1!

Any chance to get this fixed for Plone-6.0.0rc2?

Issue #510, opened 2022-07-24
theme_barceloneta: The filename, directory name, or volume label syntax is incorrect


me-kell:
* Operating System: windows

### Description

In Windows when generating a `theme_barceloneta` the path in the variable `package.dottedname` in the `mrbob.configurator` is not correctly splitted by "/" because Windows uses "\" as path separator. Subsequently `package.namespace` is wrong. Specially if the path contains dots.

The source of the problem is (https://github.com/plone/bobtemplates.plone/blob/b561328772efc5210196ba02c32de5493155654b/bobtemplates/plone/base.py#L434) 

`split("/")` should be replaced with `split(os.sep)` for it to work in Windows.

I can confirm this solves the problem in `Plone-5.2.8` and `Plone-5.2.9`.

### What I Did
```
C:\Users\vagrant\Plone-5.2.9-zeo\py-3.8.10.final.0\Scripts\activate.bat

cd C:\Users\vagrant\Plone-5.2.9-zeo\zeocluster\src

rmdir /q/s C:\Users\vagrant\Plone-5.2.9-zeo\zeocluster\src\packagename.subpackagename

copy /Y con C:\Users\vagrant\Plone-5.2.9-zeo\zeocluster\src\addon.ini
[variables]
author.name = theAuthor
author.email = info@domain.org
author.github.user = gitusername
package.description = An add-on for Plone
package.git.init = True
package.git.autocommit: True
plone.version = 5.2.8
python.version = python3.9
vscode_support = False
^Z

C:\Users\vagrant\Plone-5.2.9-zeo\zeocluster\bin\mrbob --config C:\Users\vagrant\Plone-5.2.9-zeo\zeocluster\src\addon.ini -O packagename.subpackagename bobtemplates.plone:addon

copy /Y con C:\Users\vagrant\Plone-5.2.9-zeo\zeocluster\src\theme_barceloneta.ini
[variables]
subtemplate_warning = n
theme.name = My Theme
package.git.init = False
package.git.autocommit: False
^Z

cd C:\Users\vagrant\Plone-5.2.9-zeo\zeocluster\src\packagename.subpackagename

C:\Users\vagrant\Plone-5.2.9-zeo\zeocluster\bin\mrbob --config C:\Users\vagrant\Plone-5.2.9-zeo\zeocluster\src\theme_barceloneta.ini -O packagename.subpackagename bobtemplates.plone:theme_barceloneta
```

Traceback (most recent call last):
```
[...]

PR #532, opened 2022-11-24
replace `/` with `os.path.sep`

me-kell:
Solves https://github.com/plone/bobtemplates.plone/issues/531 and https://github.com/plone/bobtemplates.plone/issues/510

### DIFF
--- a/bobtemplates/plone/base.py
+++ b/bobtemplates/plone/base.py
@@ -434,7 +434,7 @@ def base_prepare_renderer(configurator):
         raise MrBobError("No setup.py found in path!\n")
     configurator.variables["package.dottedname"] = configurator.variables[
         "package.root_folder"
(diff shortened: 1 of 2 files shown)


me-kell:
I'm afraid, I did something wrong. But I don't understand what.
I could use some help.

There is a message "Pyhon package / build (3.8) (pull request)" failing.
Which reads [Imports are incorrectly sorted and/or formatted.](https://github.com/plone/bobtemplates.plone/actions/runs/3542635445/jobs/5948345149#step:8:822)
In my commit I just replaced `/` with `os.path.sep`. See https://github.com/plone/bobtemplates.plone/pull/532/files.
I did not change any imports.

jensens:
In Plone we use isort with a rule in setup.cfg. So running isort should fix it.

Issue #533, opened 2022-11-24
Errors with `bobtemplates.plone` with version `>=6.0` caused by missing `tox.ini` file


me-kell:
### Description

Versions `6.0`, `6.1`, and `6.2` of `bobtemplates.plone` raise following errors (both in Linux and Windows):

```console
ERROR: tox config file (either pyproject.toml, tox.ini, setup.cfg) not found
Error on isort-apply:
ERROR: tox config file (either pyproject.toml, tox.ini, setup.cfg) not found
Error on black-enforce:
```
With older versions `6.0bx` or `<=5.22`  I didn't get any errors.

Older versions generated/copied a `tox.ini` file versions `>=6.0` seems not to generate a `tox.ini`.

### Probably caused by

Since version 6.0 the [MANIFEST.in](https://github.com/plone/bobtemplates.plone/blob/9bf1efef7c735ecec99d6e195aa5b10fa0e3dbfa/MANIFEST.in#L7) does `global-exclude tox.ini`.

Both [bobtemplates.plone.utils.run_isort](https://github.com/plone/bobtemplates.plone/blob/4af1ab5e02be57a1d07104616dde345e76330220/bobtemplates/plone/utils.py#L72) and [bobtemplates.plone.utils.run_black](https://github.com/plone/bobtemplates.plone/blob/4af1ab5e02be57a1d07104616dde345e76330220/bobtemplates/plone/utils.py#L91) start a subprocess calling `tox` which fails because there is no `tox.ini`.

PR #538, opened 2022-12-20
fix CI

MrTango:


.github/workflows/python-package.yml
```diff
@@ -36,7 +36,7 @@ jobs:
             - name: Install dependencies
               run: |
                   python -m pip install --upgrade pip
-                  pip install flake8 tox tox-gh-actions
+                  pip install tox-gh-actions
```
me-kell, line 39: Is this maybe related to [#533](https://github.com/plone/bobtemplates.plone/issues/533)?

mauritsvanrees:
It would be nicer to work with tox 4.
Would it be enough to replace `whitelist_externals` with `allowlist_external` in `tox.ini`?

PR #540, opened 2023-01-05
Remove effbot.org/downloads from find-links

fredvd:
No longer online. Fredrik Lundh passed away in 2021.  IIRC this was added for some special packages needed for Plone 3 & 4. Nowadays everything is on PyPI.

### DIFF
--- a/bobtemplates/plone/buildout/base.cfg.bob
+++ b/bobtemplates/plone/buildout/base.cfg.bob
@@ -28,7 +28,6 @@ versions = versions
 # of Plone packages.
 find-links =
     http://dist.plone.org
-    http://effbot.org/downloads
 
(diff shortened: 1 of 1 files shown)


gforcada (approved):

mauritsvanrees (approved):

mauritsvanrees:
This fails due to tox 4. The initial error should be fixed by replacing `whitelist_externals` with `allowlist_external` in `tox.ini`. But @MrTango is working on a PR to fix CI: #538. So maybe more is needed.

MrTango:
now it works, the problem last time was, that the GitHub plugin for tox was not compatible with the version 4 of tox. the release was done, now it works.

PR #543, opened 2023-02-02
mockup_pattern: Fix imports of config files to point to @patternslib/dev

thet:


thet:
These fixes are super-necessary.
I'm merging.

Issue #496, opened 2022-02-07
Add a `msg` funtion to the view python class (e.g. as follows):


me-kell:
The template and view in `bobtemplates/plone/view/views` are not working together.

The template uses `<p tal:content="view/msg">this gets replaced</p>` but there is no `msg` function in the view class.

Add a `msg` funtion to the view python class (e.g. as follows):

```python
def msg(self):
    return "msg from '%s'" % {{{ view_python_class_name }}}
```

An alternative would be to remove `<p tal:content="view/msg">this gets replaced</p>` from the template.

I'd prefer adding the `msg` function.

PR #547, opened 2023-03-05
fixed the msg in view , added tal:content=view/msg expression in the …

Coder-aadarsh:
As suggested by @me-kell in issue - [Add a msg funtion to the view python class (e.g. as follows): #496](https://github.com/plone/bobtemplates.plone/issues/496#issue-1126304458)
Modified code so that - 
The msg function returns a string that includes the name of the view class; and the template file using tal:content="view/msg" expression in the <p> tag will call the msg function in the view and display the returned string on the page.

### DIFF
--- a/bobtemplates/plone/view/views/+view_python_file_name+.py.bob
+++ b/bobtemplates/plone/view/views/+view_python_file_name+.py.bob
@@ -35,6 +35,10 @@ class {{{ view_python_class_name }}}({{{ view_base_class }}}):
 {{% else %}}
         return self.index()
 {{% endif %}}
+
+    def msg(self):
+        return "msg from '{}".format(self.__class__.__name__)
+
 {{% else %}}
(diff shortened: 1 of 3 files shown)


bobtemplates/plone/view/views/+view_python_file_name+.py.bob
```diff
@@ -35,6 +35,10 @@ class {{{ view_python_class_name }}}({{{ view_base_class }}}):
 {{% else %}}
         return self.index()
 {{% endif %}}
+
+    def msg(self):
+        return "msg from '{}".format(self.__class__.__name__)
```
MrTango, line 40: it's time for f-strings ;)
```suggestion
        return f"msg from {self.__class__.__name__}"
```

MrTango (changes requested):

PR #550, opened 2023-06-16
Fix SCSS imports for Bootstrap 5.3

petschki:
see https://github.com/plone/plonetheme.barceloneta/issues/335 and https://github.com/plone/buildout.coredev/pull/864#issuecomment-1584394489


mauritsvanrees (approved):
Nice!  I like the green. :-)

mauritsvanrees:
So I created a new theme with this, and it works.

But we should document how to fix existing themes.  You have some changes in this PR that are not strictly needed for that, I think.  I tried a few things.  Can you check if the following instructions are sane?

For an existing theme where you want to keep using Bootstrap 5.2, you should simply add a pin to `package.json`.  Add `"bootstrap": "~5.2.3"` to either `devDependencies` or `dependencies`.  Both seem to work for me.  Is there a preference?

If you want to update an existing theme to use Bootstrap 5.3:

* Update the Barceloneta version in `package.json` in `dependencies`: `"@plone/plonetheme-barceloneta-base": "^3.1.3"`.
* In `styles/theme.scss` add two imports in part 3: 
  * After `variables.colors.plone` add this line: `@import "@plone/plonetheme-barceloneta-base/scss/variables.colors.dark.plone";`
  * After `bootstrap/scss/variables` add this line: `@import "bootstrap/scss/variables-dark";`
* Optionally add `"bootstrap": "~5.3.0"` to `dependencies` or `devDependencies`, if you want to avoid possible problems when Bootstrap 5.4 comes out.  Does it make sense to add this in the current PR?

Where to document this?  Maybe in several places:

* In `CHANGES.rst` here in bobtemplates.plone.  People have created a theme with this, it maybe no longer works, so they update bobtemplates and maybe even check the changelog.
* I can add it to the release notes of Plone 6.0.6, which I actually already should start planning.
* A message on community.plone.org would be good as well.

mauritsvanrees:
One final thought: if we merge this PR, then bobtemplates.plone can only create themes that use Bootstrap 5.3.  No problem, especially if we go with Bootstrap 5.3 in Plone 6.0.x already.  But I would say this deserves a feature release.

We could do a bugfix release first where we add `"bootstrap": "~5.2.3"` to `devDependencies`, so there is a version of bobtemplates that can still generate a Bootstrap 5.2 theme that works out of the box.  Then in the upgrade note we can point to this version.  Does that make sense?

petschki:
> We could do a bugfix release first where we add `"bootstrap": "~5.2.3"` to `devDependencies`, so there is a version of bobtemplates that can still generate a Bootstrap 5.2 theme that works out of the box. Then in the upgrade note we can point to this version. Does that make sense?

This definitely makes sense and is the right way to go. I'll prepare a PR for `6.2.7` release and after the release we can create a bugfix branch.

/cc @MrTango

petschki:
just found out that `@plone/plonetheme-barceloneta-base=3.0.2` will pull Bootstrap 5.3 because of a wrong version pin (https://github.com/plone/plonetheme.barceloneta/pull/336/commits/6d939d6521188c8850eb26ef7a0763d003093552#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R47). For completeness sake I'll fix that in `3.0.3` and will pin that here in `6.2.x` ... then we're safe.

PR #556, opened 2023-07-19
use portal_url to root links

erral:
See https://github.com/plone/plonetheme.barceloneta/pull/341


libargutxi (approved):

MrTango (approved):
looks good, but a changelog entry would be nice ;)

PR #568, opened 2024-10-05
allow overriding package.dottedname through a configuration file

erral:
Useful to be able to override some settings when creating packages, such as when using from cookieplone to create Plone  projects.

see https://github.com/plone/cookieplone-templates/pull/63


stevepiercy:
@erral would you please add a changelog entry?

erral:
done

MrTango (approved):

PR #571, opened 2024-10-17
fix test-coverage bash script

mamico:
as it is now, in my tests, the bash script never fails, so using bin/test-coverage in an CI never shows broken tests


MrTango:
interesting, i never used that.

MrTango:
can you please add a changelog entry

MrTango (approved):

PR #580, opened 2025-04-01
Cookieplone fixes

MrTango:


erral:
These changes broke the way we had previously overriding the `package.dottedname` in the `.mrbob.ini` file:

https://github.com/plone/bobtemplates.plone/pull/568

bobtemplates/plone/base.py
```diff
@@ -226,6 +229,8 @@ def set_global_vars(configurator):
         version = bob_config.version
     configurator.variables["plone.version"] = version
     set_plone_version_variables(configurator)
+    if hasattr(bob_config, "package_dottedname"):
```
erral, line 232: This needs to check that bob_config.package_dottedname is not empty, because if you set it for instance in `.mrbob.ini` and not in `bobtemplate.cfg` for instance, bob_config.package_dottedname exists but it is None so None is set in the variable

PR #581, opened 2025-04-02
allow package structures without package_dir (src-folder)

MrTango:


erral:
These changes break current cookieplone generated structures because they don't have `setup.py` anymore.

PR #584, opened 2025-04-09
fix src packages dir detection for cookieplone

MrTango:


erral:
Tested in a recently created project with cookieplone and works as expected.

erral:
But then we need to remove the `.mrbob.ini` file from the cookieplone generated structure and add `package.dottedname` in the `bobtemplate.cfg`

MrTango:
yes, .mrbob.ini does not make much sense there anyway, as you would always have to point to it when running mr.bob.
It's not picked up by it self.
The only place where that happens is the one you have in your home directory.
So in older projects the place was bobtemplate.cfg and in upcoming versions it will be pyproject.toml

erral:
> yes, .mrbob.ini does not make much sense there anyway, as you would always have to point to it when running mr.bob. It's not picked up by it self. The only place where that happens is the one you have in your home directory. So in older projects the place was bobtemplate.cfg and in upcoming versions it will be pyproject.toml

Yeah, that's why I added the `make add` command, which adds the `-b .mrbob.ini` parameter to plonecli to include the specific configurations.

When I did that, using a custom .mrbob.ini file was the only way to customize the options for plonecli.

PR #586, opened 2025-06-10
feat: allow overriding the browserlayer

erral:
For full compatibility with cookieplone, we need to be able to override the BrowserLayer, we can pass it in the `bobtemplate.cfg` file like the `package.dottedname.

See https://github.com/plone/cookieplone-templates/pull/248 for changes in cookieplone

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,7 +9,7 @@
 root = true
 
 
-[*]  # For All Files
+[*]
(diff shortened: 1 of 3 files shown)


davisagli (approved):

davisagli:
There's a failing test. It appears to be already failing on main...

